PR #25, opened 2023-08-08
Add github workflows

lzivadinovic:
I've created two workflows
### pr-workflow 
This workflow is triggered when PR is created against main branch. It builds app and uploads debug apk to artifacts.

### tag-release workflow
This workflow is triggered when you push tag to github. It will build app, sign it , upload artifacts and create github release with that signed .apk file. You need to add few secrets to github actions settings and elevate actions GITHUB_TOKEN permission to write so it can create actual release.


AleksandarIlic (approved):
Approved. Please merge it with SQUASH & MERGE.

PR #31, opened 2023-08-16
Implement zapping

nikolalukovic:


app/schemas/net.primal.android.db.PrimalDatabase/5.json
AleksandarIlic, line 1: Revert changes on this scheme file since we already have new scheme version. This happens because we usually upgrade the scheme version after changing the entities.

app/src/main/kotlin/net/primal/android/discuss/feed/FeedViewModel.kt
```diff
@@ -171,7 +175,21 @@ class FeedViewModel @Inject constructor(
     }
 
     private fun zapPost(zapAction: UiEvent.ZapAction) = viewModelScope.launch {
+        try {
+            if (zapAction.postAuthorLightningAddress !== null) {
```
AleksandarIlic, line 179: If `postAuthorLightningAddress` is missing this means we can't zap the user, right? We should probably show some kind of message to the user that the target user can not be zapped because.

app/src/main/kotlin/net/primal/android/explore/feed/ExploreFeedViewModel.kt
```diff
@@ -121,6 +124,21 @@ class ExploreFeedViewModel @Inject constructor(
...
+                comment = zapAction.zapDescription ?: "",
+                amount = zapAction.zapAmount ?: 42,
+                target = ZapTarget.Note(
+                    id = zapAction.postId,
+                    authorPubkey = zapAction.postAuthorId,
+                    authorLightningAddress = zapAction.postAuthorLightningAddress
+                ),
+                relays = activeAccountStore.activeUserAccount().relays
+            )
+        } catch (error: IllegalArgumentException) {
```
AleksandarIlic, line 138: What is this for?

app/src/main/kotlin/net/primal/android/networking/relays/NostrWalletRelay.kt
```diff
@@ -0,0 +1,144 @@
...
+
+@Singleton
+class NostrWalletRelayPool @Inject constructor(
+    private val okHttpClient: OkHttpClient,
+    private val activeAccountStore: ActiveAccountStore,
+) {
+    private val scope = CoroutineScope(Dispatchers.IO)
+
+    private val poolMutex = Mutex()
+    private var walletConnectPool = listOf<NostrSocketClient>()
```
AleksandarIlic, line 41: Do we really need the pool for the wallet stuff? I thought we have to talk only to one relay for the wallet stuff?

app/src/main/kotlin/net/primal/android/networking/relays/RelayPool.kt
```diff
@@ -43,12 +44,43 @@ class RelayPool @Inject constructor(
         observeActiveAccount()
     }
 
+    @OptIn(FlowPreview::class)
```
AleksandarIlic, line 47: This changes seems outdated. Did you forgot to revert or?

app/src/main/kotlin/net/primal/android/nostr/ext/MetadataEvents.kt
```diff
@@ -12,13 +12,15 @@ fun List<NostrEvent>.mapAsProfileMetadataPO() = map { it.asProfileMetadataPO() }
...
     val metadata = NostrJson.decodeFromStringOrNull<ContentMetadata>(this.content)
+
     return ProfileMetadata(
         eventId = this.id,
         ownerId = this.pubKey,
         createdAt = this.createdAt,
         raw = NostrJson.encodeToString(this),
         handle = metadata?.name,
         internetIdentifier = metadata?.nip05,
+        lightningAddress = metadata?.lud16,
```
AleksandarIlic, line 23: What about lud06?

app/src/main/kotlin/net/primal/android/nostr/ext/Tags.kt
```diff
@@ -135,3 +136,27 @@ fun String.parseHashtagTags(): List<JsonArray> {
     }
     return tags.toList()
 }
+
+fun ZapTarget.toTags(): List<JsonArray> {
```
AleksandarIlic, line 140: Can we reuse the ext funs for the creation of `p` and `e` tags? Ideally every tag should be created only at one place and then reused everywhere.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
```diff
@@ -99,4 +113,44 @@ class NostrNotary @Inject constructor(
...
+            content = comment,
+            tags = tags
+        ).signOrThrow(nsec = findNsecOrThrow(userId))
+    }
+
+    fun signWalletInvoiceRequestNostrEvent(
+        request: WalletRequest<PayInvoiceRequest>,
+        nwc: NostrWallet
+    ): NostrEvent {
+        val tags = listOf(buildJsonArray {
```
AleksandarIlic, line 141: We have ext functions like this: `String.asEventIdTag()`,`String.asPubkeyTag()`.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrUnsignedEventExt.kt
```diff
@@ -27,6 +28,22 @@ fun NostrUnsignedEvent.signOrThrow(nsec: String): NostrEvent {
     )
 }
 
+fun NostrUnsignedEvent.signOrThrow(hexPrivkey: ByteArray): NostrEvent {
```
AleksandarIlic, line 31: We should do signing only at one place. All other functions should use that function. 

If doesn't make sense to convert the `ByteArray` to `nsec` format, then you can expose multiple `signOrThrow` with different inputs which all delegate the signing to just one function.

app/src/main/kotlin/net/primal/android/nostr/repository/ZapRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val activeAccountStore: ActiveAccountStore,
+    private val nostrWalletRelayPool: NostrWalletRelayPool
+) {
+    suspend fun zap(
+        comment: String = "",
+        amount: Int = 42,
+        target: ZapTarget,
+        relays: List<Relay>
+    ) {
+        val wallet = activeAccountStore.activeUserAccount().nostrWallet ?: throw RuntimeException()
```
AleksandarIlic, line 23: We should probably get rid of `RuntimeException` and throw something that we defined in `ZapRepository` which can be caught in then appropriate action can be taken.

app/src/main/kotlin/net/primal/android/nostr/repository/ZapRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        val lightningAddress = when (target) {
+            is ZapTarget.Note -> target.authorLightningAddress
+            is ZapTarget.Profile -> target.lightningAddress
+        }
+
+        requireNotNull(lightningAddress) { "Lightning address of the recipient is missing" }
```
AleksandarIlic, line 30: We could probably delegate this checks to the caller which shouldn't be able to call `zap` without `lightningAddress`.

app/src/main/kotlin/net/primal/android/user/domain/NostrWalletConnectParser.kt
```diff
@@ -15,12 +18,19 @@ fun String.parseNWCUrl(): NostrWallet {
...
     }
 
-    return NostrWallet(pubkey = pubkey, lud16 = lud16, relayUrl = relay, secret = secret)
+    val keypairPubkey = CryptoUtils.publicKeyCreate(Hex.decode(keypairPrivkey)).toHex()
+
+    return NostrWallet(
+        pubkey = pubkey,
+        lud16 = lud16,
+        relayUrl = relay,
+        keypair = NostrWalletKeypair(privkey = keypairPrivkey, pubkey = keypairPubkey)
```
AleksandarIlic, line 32: Can we store just one `secret` and one `public key`? Or there are multiple pubkeys here?

`NostrWalletKeypair` can also be created whenever we needed, right? We don't have to create it when parsing?

AleksandarIlic (commented):
Great progress! Just few minor comments.

app/schemas/net.primal.android.db.PrimalDatabase/5.json
nikolalukovic: will do

app/src/main/kotlin/net/primal/android/discuss/feed/FeedViewModel.kt
nikolalukovic: agreed, is there already an example on how to do that in the codebase?

app/src/main/kotlin/net/primal/android/explore/feed/ExploreFeedViewModel.kt
nikolalukovic: I drilled nullable lightning address further, but this will be removed once we show a message to user that lightning address is missing.

app/src/main/kotlin/net/primal/android/networking/relays/NostrWalletRelay.kt
nikolalukovic: I'm not sure how it will work once user is able to connect to multiple NWC APIs, so I left it like this until that gets cleared out since it was easier to just use existing API.

app/src/main/kotlin/net/primal/android/networking/relays/RelayPool.kt
nikolalukovic: Forgot to revert.

app/src/main/kotlin/net/primal/android/nostr/ext/MetadataEvents.kt
nikolalukovic: lud06 is outdated and not in use pretty much anywhere

app/src/main/kotlin/net/primal/android/nostr/ext/Tags.kt
nikolalukovic: Will do.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
nikolalukovic: Didn't see those. I'll use them 👍🏻

app/src/main/kotlin/net/primal/android/discuss/feed/FeedViewModel.kt
AleksandarIlic: When we post new note we handle some errors and show them to the user.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrUnsignedEventExt.kt
nikolalukovic: I've made this one to avoid additional conversions and allocations like this: hex string -> nsec string -> hex string -> byte array since hex byte array is used in the end anyways.

app/src/main/kotlin/net/primal/android/nostr/repository/ZapRepository.kt
nikolalukovic: Yes, that was left as an interim solution.

app/src/main/kotlin/net/primal/android/nostr/repository/ZapRepository.kt
nikolalukovic: Yes as we agreed in the previous comment

app/src/main/kotlin/net/primal/android/user/domain/NostrWalletConnectParser.kt
nikolalukovic: No. Pubkey from host and pubkey received from the secret private key are different and used for different purposes. I've immedietely created `NostrWalletKeypair` here to avoid doing that every time we zap.

app/src/main/kotlin/net/primal/android/networking/relays/NostrWalletRelay.kt
AleksandarIlic: I'm not sure we'll support simultaneous connections to multiple wallets. Let's discuss that tomorrow on a daily call.

AleksandarIlic (approved):

AleksandarIlic (approved):

PR #34, opened 2023-08-22
Implement mutiny wallet and better deep link url handling

nikolalukovic:


app/src/main/kotlin/net/primal/android/core/utils/ValidationUtils.kt
```diff
@@ -9,6 +9,8 @@ fun String?.isValidNostrKey(): Boolean {
 
     return if (startsWith("nsec")) {
         this.isValidNsec()
+    } else if (startsWith("npub")) {
+        this.isValidNpub()
```
AleksandarIlic, line 13: This will break login because it will allow logging in with npub as a private key. The fun name is a bit misleading. You can rename it to `isValidNostrPrivateKey` if you want.

AleksandarIlic (commented):
Looks good. Please fix the mentioned side-effect and we are good to go.

app/src/main/kotlin/net/primal/android/core/utils/ValidationUtils.kt
nikolalukovic: Pushed new commit with changes

AleksandarIlic (approved):

PR #40, opened 2023-09-13
Create account feature

nikolalukovic:


app/build.gradle.kts
```diff
@@ -57,8 +57,8 @@ android {
         applicationId = "net.primal.android"
         minSdk = 26
         targetSdk = 34
-        versionCode = 33
-        versionName = "0.15.2"
+        versionCode = 34
+        versionName = "0.16.0-feature-create-account-pre-release"
```
AleksandarIlic, line 61: In the future please don't add descriptive suffix. Changes can be described in the RELEASE_NOTES.md.

app/src/main/kotlin/net/primal/android/auth/create/api/model/RecommendedFollowsResponse.kt
```diff
@@ -0,0 +1,28 @@
...
+import net.primal.android.nostr.model.NostrEventKind
+
+@Serializable
+data class Suggestion(val name: String, val pubkey: String)
+
+@Serializable
+data class SuggestionGroup(val group: String, val members: List<Suggestion>)
+
+@Serializable
+data class Metadata(
```
AleksandarIlic, line 14: Why not using `NostrEvent`?

app/src/main/kotlin/net/primal/android/auth/create/api/model/CreateNostrProfileMetadata.kt
```diff
@@ -0,0 +1,17 @@
+package net.primal.android.auth.create.api.model
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class CreateNostrProfileMetadata(
```
AleksandarIlic, line 7: We already have `ContentMetadata` in `nostr.model.content` package. Can it be reused?

app/src/main/kotlin/net/primal/android/auth/create/CreateContract.kt
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    sealed class UiEvent {
+        data object GoToProfilePreviewStepEvent : UiEvent()
+        data object GoToNostrCreatedStepEvent : UiEvent()
+        data object GoToFollowContactsStepEvent : UiEvent()
+        data object GoBack : UiEvent()
+        data object FinishEvent : UiEvent()
+        data class AvatarUriChangedEvent(val avatarUri: Uri?) : UiEvent()
```
AleksandarIlic, line 50: Maybe you can store the states inside of the screen composable and only on Next update the `ViewModel`. Or maybe we could also keep everything in the screen and pass the event to view model on final action. The current approach has too many small events which just update the state which probably could be kept inside of the screen.

`mutableStateOf` can be used in Composables with `remember` functions. `rememberSaveable` is good for saving the state across configuration changes as well which we should do in this case.

app/src/main/kotlin/net/primal/android/core/compose/PrimalLoadingSpinner.kt
```diff
@@ -16,7 +16,7 @@ import com.airbnb.lottie.compose.rememberLottieComposition
 import net.primal.android.R
 
 @Composable
-fun PrimalLoadingSinner(
+fun PrimalLoadingSpinner(
```
AleksandarIlic, line 19: Loading sins have been forgiven. Great catch.

app/src/main/kotlin/net/primal/android/networking/primal/PrimalClient.kt
```diff
@@ -4,19 +4,16 @@ import kotlinx.coroutines.delay
...
 import net.primal.android.networking.sockets.errors.WssException
 import net.primal.android.networking.sockets.filterBySubscriptionId
 import java.util.UUID
 import javax.inject.Inject
-import javax.inject.Singleton
 
-@Singleton
-class PrimalApiClient @Inject constructor(
-    @PrimalSocketClient private val socketClient: NostrSocketClient,
+class PrimalClient @Inject constructor(
```
AleksandarIlic, line 15: Why changing the name?

app/src/main/kotlin/net/primal/android/networking/di/SocketModule.kt
```diff
@@ -42,4 +61,22 @@ object SocketModule {
...
+    @Named("Api")
+    fun providesPrimalApiClient(
+        @PrimalSocketClient socketClient: NostrSocketClient
+    ) = PrimalClient(
+        socketClient = socketClient
+    )
+
+    @Provides
+    @Singleton
+    @Named("Upload")
```
AleksandarIlic, line 76: Can we use custom annotations?

app/src/main/kotlin/net/primal/android/networking/relays/RelayPool.kt
```diff
@@ -42,6 +44,7 @@ class RelayPool @Inject constructor(
 
     private var regularRelays = listOf<NostrSocketClient>()
     private var walletRelays = listOf<NostrSocketClient>()
+    private var bootstrapRelays = listOf<NostrSocketClient>()
```
AleksandarIlic, line 47: This relays stuff needs to be refactored. We are just adding to much complexity here and needs to be simplified somehow. I don't have any ideas at the moment, I'll have to think about it how to approach this.

The more complexity goes up the more chances are that something can be broken, misunderstood or misused in the future. That is the main concern.

app/src/main/kotlin/net/primal/android/networking/sockets/NostrSocketClient.kt
```diff
@@ -15,8 +15,9 @@ import okhttp3.WebSocket
 import okhttp3.WebSocketListener
 import timber.log.Timber
 import java.util.UUID
+import javax.inject.Inject
 
-class NostrSocketClient constructor(
+class NostrSocketClient @Inject constructor(
```
AleksandarIlic, line 20: Why add? This was intentionally left so we couldn't instantiate this class with DI.

app/src/main/kotlin/net/primal/android/nostr/model/content/ContentMetadata.kt
```diff
@@ -14,3 +16,11 @@ data class ContentMetadata(
     val banner: String? = null,
     val website: String? = null,
 )
+
+fun ContentMetadata.userNameUiFriendly(pubkey: String): String {
```
AleksandarIlic, line 20: This introduces the same business logic in two places now. It already exists in `ProfileMetadata`. The business logic of user|author friendly names should be only in one place.

The solution is probably that both `ContentMetadata` and `ProfileMetadata` use that business logic.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
```diff
@@ -156,4 +158,63 @@ class NostrNotary @Inject constructor(
...
+    ): NostrEvent {
+        return NostrUnsignedEvent(
+            pubKey = pubkey,
+            kind = NostrEventKind.Metadata.value,
+            content = json.encodeToString(metadata),
+            tags = emptyList()
+        ).signOrThrow(hexPrivateKey = Hex.decode(privkey))
+    }
+
+    fun signFirstContactNostrEvent(
```
AleksandarIlic, line 188: Why can't we reuse the default one? The same question applies for `signInitialContactsNostrEvent`?

app/src/main/kotlin/net/primal/android/auth/create/compose/CreateScreen.kt
```diff
@@ -0,0 +1,311 @@
...
+import net.primal.android.auth.create.compose.steps.ProfilePreviewStep
+import net.primal.android.core.compose.PrimalTopAppBar
+import net.primal.android.core.compose.button.PrimalLoadingButton
+import net.primal.android.core.compose.icons.PrimalIcons
+import net.primal.android.core.compose.icons.primaliconpack.ArrowBack
+import net.primal.android.nostr.model.content.ContentMetadata
+import net.primal.android.theme.PrimalTheme
+
+@Composable
+fun CreateScreen(
```
AleksandarIlic, line 59: I have to look into this code from IDE in more details. Can't do it today.

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/auth/create/api/model/RecommendedFollowsResponse.kt
nikolalukovic: It didn't cross my mind to use it, lul. I'll switch to it.

app/src/main/kotlin/net/primal/android/auth/create/api/model/CreateNostrProfileMetadata.kt
nikolalukovic: It didn't cross my mind to use it, lul. I'll switch to it.

app/src/main/kotlin/net/primal/android/networking/sockets/NostrSocketClient.kt
nikolalukovic: Forgot to remove it. Will remove.

app/src/main/kotlin/net/primal/android/networking/primal/PrimalClient.kt
nikolalukovic: The reasoning behind it was that we're using two instances of it. One to consume the API and the other to consume the UPLOAD

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
nikolalukovic: We're not logged in when this has to be used.

app/src/main/kotlin/net/primal/android/auth/create/CreateContract.kt
nikolalukovic: 👍🏻 I'll change it

app/src/main/kotlin/net/primal/android/nostr/model/content/ContentMetadata.kt
nikolalukovic: It feels wrong to use `ProfileMetadata` considering it's a domain model bound to db. While using `ContentMetadata` here feels more natural considering it's just a transient API consumption model.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
AleksandarIlic: Why can't we login the user first and then perform all other operations? In case something goes wrong, we can rollback/logout or provide some other fallback solution.

app/src/main/kotlin/net/primal/android/nostr/model/content/ContentMetadata.kt
AleksandarIlic: I agree. If the business logic is in more than one place it can cause maintenance problems in the future.

We should put it in one place and make some wrappers/helpers around ti for `ContentMetadata` and `ProfileMetadata`.

app/src/main/kotlin/net/primal/android/networking/primal/PrimalClient.kt
AleksandarIlic: Yes, but upload is still some kind of API.

AleksandarIlic (approved):

PR #44, opened 2023-09-25
Initial edit account view

nikolalukovic:


app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -274,6 +278,8 @@ fun PrimalAppNavigation() {
                 navController = navController,
             )
 
+            editProfile(route = "edit-profile", navController = navController)
```
AleksandarIlic, line 281: edit_profile

app/src/main/kotlin/net/primal/android/profile/details/ProfileContract.kt
```diff
@@ -11,6 +11,7 @@ interface ProfileContract {
     data class UiState(
         val profileId: String,
         val isProfileFollowed: Boolean,
+        val isCurrentUser: Boolean,
```
AleksandarIlic, line 14: isActiveUser

app/src/main/kotlin/net/primal/android/profile/details/ProfileScreen.kt
```diff
@@ -601,9 +613,17 @@ private fun ProfileActions(
...
+            when (isFollowed) {
+                true -> UnfollowButton(onClick = onUnfollow)
+                false -> FollowButton(onClick = onFollow)
+            }
+        }
+        Spacer(modifier = Modifier.width(8.dp))
+        if (isCurrentUser) {
+            EditProfileButton(onClick = {
+                onEditProfileClick()
+            })
```
AleksandarIlic, line 626: You either need to show Follow/Unfollow buton or Edit Profile. If/else should be enough.

app/src/main/kotlin/net/primal/android/profile/edit/EditProfileScreen.kt
```diff
@@ -0,0 +1,226 @@
...
+    state: EditProfileContract.UiState,
+    eventPublisher: (EditProfileContract.UiEvent) -> Unit,
+    paddingValues: PaddingValues,
+    modifier: Modifier = Modifier
+) {
+    Column(
+        modifier = modifier
+            .fillMaxSize()
+            .systemBarsPadding()
+            .navigationBarsPadding()
```
AleksandarIlic, line 96: Remove systemBarsPadding() and navigationsBarsPadding(). It's causing padding issues.

app/src/main/kotlin/net/primal/android/profile/edit/EditProfileContract.kt
```diff
@@ -0,0 +1,50 @@
...
+        data class AboutMeChangedEvent(val aboutMe: String) : UiEvent()
+        data class LightningAddressChangedEvent(val lightningAddress: String) : UiEvent()
+        data class Nip05IdentifierChangedEvent(val nip05Identifier: String) : UiEvent()
+        data class AvatarUriChangedEvent(val avatarUri: Uri?) : UiEvent()
+        data class BannerUriChangedEvent(val bannerUri: Uri?) : UiEvent()
+        data object SaveProfileEvent : UiEvent()
+    }
+}
+
+fun EditProfileContract.UiState.toProfileMetadata(): ProfileMetadata {
```
AleksandarIlic, line 39: This shouldn't be in Contract. Move it to viewModel.

app/src/main/kotlin/net/primal/android/profile/edit/EditProfileContract.kt
```diff
@@ -0,0 +1,50 @@
...
+    return ProfileMetadata(
+        displayName = this.displayName,
+        handle = this.name,
+        website = this.website,
+        about = this.aboutMe,
+        lightningAddress = this.lightningAddress,
+        nostrVerification = this.nip05Identifier,
+        picture = this.avatarUri,
+        banner = this.bannerUri
+    )
```
AleksandarIlic, line 49: We need to make sure that we don't send empty strings. You can use `.ifEmpty { null }` extension function to return null if the fields are not nullables.

app/src/main/kotlin/net/primal/android/profile/edit/EditProfileScreen.kt
```diff
@@ -0,0 +1,226 @@
...
+    eventPublisher: (EditProfileContract.UiEvent) -> Unit,
+    onClose: () -> Unit
+) {
+    val snackbarHostState = remember { SnackbarHostState() }
+    ErrorHandler(error = state.error, snackbarHostState = snackbarHostState)
+
+    Scaffold(
+        topBar = {
+            PrimalTopAppBar(
+                title = "Edit Profile",
```
AleksandarIlic, line 68: Move to strings.xml.

app/src/main/kotlin/net/primal/android/profile/edit/EditProfileViewModel.kt
```diff
@@ -0,0 +1,138 @@
...
+            setState {
+                copy(
+                    displayName = it.metadata?.displayName ?: "",
+                    name = it.metadata?.handle ?: "",
+                    aboutMe = it.metadata?.about ?: "",
+                    website = it.metadata?.website ?: "",
+                    lightningAddress = it.metadata?.lightningAddress ?: "",
+                    nip05Identifier = it.metadata?.internetIdentifier ?: "",
+                    bannerUri = it.metadata?.banner?.toUri(),
+                    avatarUri = it.metadata?.picture?.toUri()
```
AleksandarIlic, line 89: Regarding the situation where we can have both local and remote images.

We could probably update `ProfileMetadata` domain class to have `localPictureUri`, `localBannerUri`, `remotePictureUrl` and `remoteBannerUrl`, and then handle all four potential use cases where appropriate. 

You will also need to update the `ProfileRepository` to fallback to `remote` urls if there are no local one provided.

AleksandarIlic (changes requested):
Overall looks very good. Just few minor things.

app/src/main/kotlin/net/primal/android/profile/edit/EditProfileContract.kt
nikolalukovic: It's a `Uri`. I don't think `Uri` could be created with an empty string. Both are already nullable.

AleksandarIlic: I was referring to String values. For example if there was no nip05 value in metadata and no nip05 value in editor, we would submit nostr event with nip05="".

PR #58, opened 2023-11-22
fix error when decrypting messages

greenart7c3:


app/src/main/kotlin/net/primal/android/nostr/ext/DirectMessageEvents.kt
```diff
@@ -54,6 +54,8 @@ fun NostrEvent.mapAsMessageDataPO(userId: String, nsec: String): DirectMessageDa
     } catch (error: GeneralSecurityException) {
         Timber.w(NostrJson.encodeToString(this.toJsonObject()))
         return null
+    } catch (error: Exception) {
```
AleksandarIlic, line 57: @greenart7c3 Do you know what kind of exception is thrown?

greenart7c3: it's a IndexOutOfBoundsException. Should i catch that instead of Exception?

AleksandarIlic: Yes, please. That would be better. And you can also instead of `String` do the same code as for `GeneralSecurityException`.

I just looked and this exception is thrown with `CryptoUtils.getSharedSecret()` probably. Were you ever able to see DMs in the app?

greenart7c3: The event is a test event without encryption that has the content "Could not decrypt the message".
The error happens in CryptoUtils line 81
val iv = parts[1].run { Base64.decode(this) }

greenart7c3: Maybe it's better to return the content when the event has no "?iv=".
I was never able to open the dms screen

AleksandarIlic: Aahh.. yeah, I see, there is no `?iv=` in the message. Thanks for the insights.

AleksandarIlic (approved):

PR #72, opened 2024-01-31
Fix default zap amount and message not used

IzzySoft:
Oops… my bad, must have gotten distracted: one directory level was missing. So this fixes #69 (apologies for the lapsus!)

Not sure why your other commit got included here as I've updated/pulled `main` before creating this branch :man_shrugging: Feel free to ignore/close this PR and move the files yourself if you prefer…


AleksandarIlic (approved):

AleksandarIlic:
Thanks! I used the force with push few days ago and now ghost commits are haunting me.

IzzySoft:
> I used the force with push few days ago

Ugh… on main? :scream: Yeah, funny side-effects on all forks then. On a branch before merging (and removing) it force push should be OK, but better avoid it on main or branches still worked on…

Thanks for merging! I'll remove my fork now (can always create a fresh one), so only 5 others remain which might be affected. Wishing you best luck to get rid of the ghosts!

Issue #98, opened 2024-03-03
Capital letter doesn't show the accent in "E" letter.


bitcoinuser:
Hi,

Capital letter doesn't show the accent in "E" letter. Should be "É"

See the video:

https://github.com/PrimalHQ/primal-android-app/assets/1178856/8556c274-9d5a-47f1-b908-aa64d7c362a3

davotoula:
I can't reproduce it. What device are you using please?

bitcoinuser:
> I can't reproduce it. What device are you using please?

Xiaomi Redmi Note 9s. 

Obs. The accent doesn't appear when writing a note, but I can see the same note with accent in other nostr clients.

davotoula:
It could still be a bug in the primal input field for your device. Let me see if I can reproduce it!

PR #101, opened 2024-03-04
added font family to note edit text field

davotoula:
This is odd... Currently AppTheme.typography.bodyMedium is used in OutlinedTextField and it has font family set as SansSerif... but I had to manually declare it.

Seems to have fixed the cut of accent issue.

Tested on Pixel 6 pro emulator (had the issue) and Pixel 7 Pro (didn't have the issue)

#98


AleksandarIlic:
This changes the app's default font when you set the font family to SansSerif. We can't use it as a solution if it changes the font.

davotoula:
Good catch. I didn't even notice the change in font! 😂

PR #107, opened 2024-03-08
cleaner code and added tests

davotoula:
replace String length with indices
use indexing operator instead of get call
added tests


davotoula:
Ooo... checks are already working! 👍

AleksandarIlic (approved):

Issue #104, opened 2024-03-05
Links in profiles that don't start with https:// don't open


rottenwheel:
When tapping on links such as designwithcarter.com vs. https://rottenwheel.com/, former displays a pop-up message: 'App not found that could open designwithcarter.com'

On latest Primal Android release go to Carter's profile (npub1p3j8q3u4l0lddrhj9yumju0lnrj0kdhtalgh96l86a8gkln5kj6qdwnh6z) and try opening the link in bio, you'll get the pop-up message seen in screenshot below.

![image](https://github.com/PrimalHQ/primal-android-app/assets/92872541/aa2284dc-bf26-4356-9839-cb978c1b4877)

On the other hand, if you pull up my profile (npub164xgt7nysd3euvrdk0h6p7xxhulhnjnpmu0utzau86wf6859qn6srac0qx), that contains the https://, it opens the link in your system's browser just fine.

AleksandarIlic:
This was fixed in 0.96. Please retest.

PR #108, opened 2024-03-08
104 add http protocol if user website is missing it

davotoula:
Fix for #104


app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsScreen.kt
```diff
@@ -715,17 +715,25 @@ private fun UserProfileDetails(
         }
 
         if (state.profileDetails?.website?.isNotEmpty() == true) {
+
+            // add protocol if user website is missing it
+            // use http protocol since many websites redirect to https if available
+            val websiteWithProtocol = if (state.profileDetails.website.startsWith("http", true))
```
AleksandarIlic, line 721: Please use named arguments in startsWith(). It will be more readable when looking at the code outside of the IDE.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsScreen.kt
```diff
@@ -715,17 +715,25 @@ private fun UserProfileDetails(
...
+                state.profileDetails.website
+            else
+                "http://" + state.profileDetails.website
```
AleksandarIlic, line 724: I would prefer "https://". 

And please remove the comment above, since the code will be self explanatory when we use `https`.

AleksandarIlic (changes requested):
Thanks for the PR. I had just two minor comments.

davotoula:
changes done

AleksandarIlic (approved):

PR #109, opened 2024-03-09
added Swedish translation (and lint disable MissingTranslations)

davotoula:
Screenshot_20240309_132313
Screenshot_20240309_132444
Screenshot_20240309_132355


davotoula:
Change android system language to view app in Swedish

AleksandarIlic:
Tack för PR. Unfortunately this can't be merged right now since many of the messages are coming from the backend as english messages and they would need to be i18n-ed on the backend as well and there is not yet an infrastructure for that.

davotoula:
Not a problem. I thought a UI in swedish would be better than nothing but I understand your concern about consistency!

Issue #92, opened 2024-02-29
Profile appear as nostr:nprofile... but should appear @adam3us


bitcoinuser:
Hi,

Profile appear as nostr:nprofile... but should appear @adam3us. Only appear adam3us if we click on the note. See the video.

https://github.com/PrimalHQ/primal-android-app/assets/1178856/c4a77a2c-b423-404c-929a-c4aa5a5dfde1

PR #112, opened 2024-03-13
92 fix for mention appear as nostr nprofile instead of at username 

davotoula:
ellipsize as a last step to prevent truncation of nostr uri


davotoula:
Fix for #92 

After:
Screenshot_20240313_193204

PR #118, opened 2024-04-09
Implement bookmarks

AleksandarIlic:


AleksandarIlic:
Tests are false positive and in need of refactor. Ran them locally with success.

PR #168, opened 2024-09-09
Feature: advanced search

markocic:


app/src/main/kotlin/net/primal/android/explore/asearch/ui/SingleChoicePicker.kt
```diff
@@ -0,0 +1,83 @@
...
+) {
+    val scope = rememberCoroutineScope()
+
+    ModalBottomSheet(
+        tonalElevation = 0.dp,
+        modifier = modifier,
+        onDismissRequest = onDismissRequest,
+        sheetState = sheetState,
+    ) {
+        PrimalTopAppBar(
```
AleksandarIlic, line 45: Title in all bottom sheet top bars is not properly centered. Probably because of the missing navigation icon. To avoid messing up other screens we should use `CenterAlignedTopAppBar` directly.

AleksandarIlic (changes requested):
Looks really good. Just a few changes are needed before we can merge this.

AleksandarIlic (approved):
Great job!

PR #174, opened 2024-09-16
Feature: profile details horizontal pager

markocic:


app/src/main/kotlin/net/primal/android/articles/feed/ArticleFeedList.kt
```diff
@@ -118,14 +121,15 @@ private fun ArticleFeedList(
...
+    PrimalPullToRefreshBox(
         isRefreshing = pullToRefreshing,
         onRefresh = {
             pagingItems.refresh()
             pullToRefreshing = true
             isMediatorRefreshing = true
         },
         state = pullToRefreshState,
+        enabled = pullToRefreshEnabled,
         indicator = {
```
AleksandarIlic, line 133: Can we use the default composable here (indicator)? Would that work?

AleksandarIlic (approved):

PR #178, opened 2024-09-20
Feature: restore default feeds

markocic:


app/src/main/kotlin/net/primal/android/feeds/FeedsViewModel.kt
```diff
@@ -117,10 +117,24 @@ class FeedsViewModel @AssistedInject constructor(
                     is UiEvent.UpdateFeedSpecEnabled -> {
                         updateFeedSpecEnabled(feedSpec = it.feedSpec, enabled = it.enabled)
                     }
+
+                    UiEvent.RestoreDefaultPrimalFeeds -> {
+                        setState { copy(isEditMode = false) }
```
AleksandarIlic, line 122: Disabling edit mode should happen after reset has been completed.

app/src/main/kotlin/net/primal/android/feeds/FeedsViewModel.kt
```diff
@@ -117,10 +117,24 @@ class FeedsViewModel @AssistedInject constructor(
...
+                        restoreDefaultPrimalFeeds()
+                    }
                 }
             }
         }
 
+    private fun restoreDefaultPrimalFeeds() =
+        viewModelScope.launch {
+            try {
+                feedsRepository.fetchAndPersistDefaultFeeds(FeedSpecKind.Notes)
```
AleksandarIlic, line 132: When we init a ViewModel we already fetch default feeds with a reason to check for new feeds and them at the end of the list.

Maybe we could cache the results from that previous api call and just replace them in a instant? And in case there are none, we call the api.

app/src/main/kotlin/net/primal/android/feeds/ui/FeedList.kt
```diff
@@ -161,6 +180,27 @@ fun FeedList(
                         )
                     }
                 }
+                if (isEditMode) {
+                    item {
+                        ListItem(
```
AleksandarIlic, line 185: Can we have the same ripple effect as on other list items?

AleksandarIlic (approved):

PR #182, opened 2024-09-23
Implement tabs on explore home screen

markocic:


app/src/main/res/values/strings.xml
```diff
@@ -237,6 +237,11 @@
     <string name="explore_enter_query">Enter text to search</string>
     <string name="explore_trending_hashtags_no_content">Trending hashtags will appear here.</string>
     <string name="explore_fallback_title">Feed</string>
+    <string name="explore_feeds_tab">FEEDS</string>
+    <string name="explore_people_tab">PEOPLE</string>
+    <string name="explore_zaps_tab">ZAPS</string>
+    <string name="explore_media_tab">MEDIA</string>
+    <string name="explore_topics_tab">TOPICS</string>
```
AleksandarIlic, line 244: Usually in Strings we put "regular" strings (in this example it would be "Topics"), and if we want any string emphasized with caps lock, we can do in UI something like `String.toUpperCase()`.

AleksandarIlic (approved):
Just a minor nitpick. Everything looks good.

PR #201, opened 2024-10-20
WIP: Fix image aspect ratio on embedded notes

markocic:


AleksandarIlic:
We can not do it like this because feeds could "jump" once image is loaded.

PR #211, opened 2024-10-22
Implement profile search

markocic:


app/src/main/kotlin/net/primal/android/navigation/NavigationArguments.kt
```diff
@@ -30,6 +30,15 @@ inline val SavedStateHandle.renderType: String
...
 inline val SavedStateHandle.initialQuery: String? get() = get(INITIAL_QUERY)
 
+const val POSTED_BY = "postedBy"
+inline val SavedStateHandle.postedBy: List<String>?
+    get() = get<String>(POSTED_BY)
+        ?.trim()?.removeSurrounding(prefix = "[", suffix = "]")
+        ?.splitToSequence(",")
+        ?.map { it.trim() }
+        ?.filter { it.isNotEmpty() }
+        ?.toList()
```
AleksandarIlic, line 40: Could we try?
```
get() = get<String>(POSTED_BY)?.let {
    NostrJson.decodeFromString(it)
}
```

We are already using `NostrJson` and Base64 for complex data transfers. This might even work without changing the sender part since it should be already a valid json array of strings.

AleksandarIlic (changes requested):
Looks good. Maybe there is a better way for transferring complex objects between screens.

AleksandarIlic (approved):

PR #215, opened 2024-10-26
Feature: coil avatar cache

markocic:


app/src/main/kotlin/net/primal/android/core/images/PrimalImageLoaderFactory.kt
```diff
@@ -7,13 +7,14 @@ import coil.ImageLoaderFactory
...
 import coil.decode.VideoFrameDecoder
+import coil.disk.DiskCache
 import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class PrimalImageLoaderFactory @Inject constructor(
-    @ApplicationContext context: Context,
+    @ApplicationContext private val context: Context,
```
AleksandarIlic, line 17: It's not good idea to keep references to Context.

You should keep it as it used to be and use `lazy` to retrieve what you need. For example, you could add following field as:

```
private val imageCacheDir by lazy {
  context.cacheDir.resolve("image_cache")
}
```

AleksandarIlic (approved):

AleksandarIlic (approved):

PR #216, opened 2024-10-26
Implement follow approvals on every profile follow/unfollow action

markocic:


app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsContract.kt
```diff
@@ -28,6 +28,8 @@ interface ProfileDetailsContract {
         val error: ProfileError? = null,
         val zapError: UiError? = null,
         val zappingState: ZappingState = ZappingState(),
+        val shouldApproveFollow: Boolean = false,
+        val shouldApproveUnfollow: Boolean = false,
```
AleksandarIlic, line 32: Yeah, we should definitely have shared `data class` which would be reused on all contracts where follow/unfollow is possible.

AleksandarIlic (changes requested):
We need to reconsider business logic when to ask for follow/unfollow approval and improve how we keep approval data in contracts.

PR #237, opened 2024-11-28
Implement follow approvals

AleksandarIlic:


app/src/main/kotlin/net/primal/android/explore/home/people/ExplorePeopleContract.kt
```diff
@@ -10,12 +10,21 @@ interface ExplorePeopleContract {
         val people: List<ExplorePeopleData> = emptyList(),
         val userFollowing: Set<String> = emptySet(),
         val error: UiError? = null,
+        val shouldApproveFollow: Boolean = false,
+        val shouldApproveUnfollow: Boolean = false,
```
AleksandarIlic, line 14: Maybe just single field which would be data class holding profileId and what action needs to be approved.

app/src/main/kotlin/net/primal/android/explore/home/people/ExplorePeople.kt
```diff
@@ -81,6 +86,14 @@ fun ExplorePeople(
     eventPublisher: (ExplorePeopleContract.UiEvent) -> Unit,
     onProfileClick: (String) -> Unit,
 ) {
+    var lastFollowUnfollowProfileId by rememberSaveable { mutableStateOf<String?>(null) }
```
AleksandarIlic, line 89: Not sure about this approach.

Maybe ViewModel should take care of which profileId and which action needs to be approved. That way the entire logic will be centralised at one place.

app/src/main/kotlin/net/primal/android/profile/repository/ProfileRepository.kt
```diff
@@ -110,35 +110,50 @@ class ProfileRepository @Inject constructor(
...
+        val userFollowList = userAccountFetcher.fetchUserFollowListOrNull(userId = userId)
+            ?: throw FollowListNotFound()
 
-            setFollowList(
-                userId = userId,
-                contacts = userFollowList.following.reducer(),
-                content = userFollowList.followListEventContent ?: "",
-            )
+        if (userFollowList.following.isEmpty() && !forceUpdate) {
+            throw PossibleFollowListCorruption()
```
AleksandarIlic, line 145: We might need to rethink the logic here from the scratch and only one exception would be needed here

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsContract.kt
```diff
@@ -26,6 +26,8 @@ interface ProfileDetailsContract {
             ProfileFeedSpec.AuthoredMedia,
         ),
         val error: ProfileError? = null,
+        val shouldApproveFollow: Boolean = false,
+        val shouldApproveUnfollow: Boolean = false,
```
AleksandarIlic, line 30: Yeah, we should definitely have shared data class which would be reused on all contracts where follow/unfollow is possible.

AleksandarIlic (commented):
We need to reconsider business logic when to ask for follow/unfollow approval and improve how we keep approval data in contracts.

PR #240, opened 2024-11-30
Fix missing apply button on filter selection when font is enlarged

markocic:


AleksandarIlic (commented):
BottomSheet is taking full height with this approach. It turns out the `verticallScroll` modifier was on the wrong `Column`. I fixed it here: 7bc836d1d7c99d26dfd56b3ac8f1d96df7ec9b7d

PR #242, opened 2024-11-30
Integrate primal premium into profile editor

markocic:


AleksandarIlic (approved):
Good job! I moved the premium check to the top to avoid unnecessary api call for lightning address.

PR #247, opened 2024-12-09
Implement legendary profile customization on referenced articles and notes

markocic:


app/src/main/kotlin/net/primal/android/nostr/ext/NostrResources.kt
```diff
@@ -335,6 +335,7 @@ private fun takeAsReferencedNoteOrNull(
         authorAvatarCdnImage = refPostAuthor.avatarCdnImage,
         authorInternetIdentifier = refPostAuthor.internetIdentifier,
         authorLightningAddress = refPostAuthor.lightningAddress,
+        authorPrimalPremiumInfo = refPostAuthor.primalPremiumInfo,
```
AleksandarIlic, line 338: `refPostAuthor.primalPremiumInfo.legendProfile`
We only need this. Better to reduce data consumption.

AleksandarIlic (changes requested):
Minor request to minimize data consumption.

AleksandarIlic (approved):

PR #253, opened 2024-12-12
Implement hide/show highlights context option on ArticleDetailsScreen

markocic:


AleksandarIlic (approved):
All looks good! And works fast as well!

PR #254, opened 2024-12-14
Implement highlight activity bottom sheet

markocic:


AleksandarIlic (approved):
Looks and works very well.

PR #259, opened 2024-12-19
Refactor user badge component

mehmedalijaK:


app/src/main/kotlin/net/primal/android/premium/legend/custimization/LegendaryProfileCustomizationScreen.kt
```diff
@@ -145,7 +145,12 @@ fun LegendaryProfileCustomizationScreen(
                     displayName = primalName,
                     internetIdentifier = "$primalName@primal.net",
                     internetIdentifierBadgeSize = 24.dp,
-                    customBadgeStyle = if (customBadge == true) selectedStyle else null,
+                    authorLegendaryCustomization =
+                    if (customBadge == true) {
```
markocic, line 149: This could be simplified by setting `customBadge` inside `LegendaryCustomization`.

```kt
authorLegendaryCustomization = LegendaryCustomization(customBadge = customBadge, legendaryStyle = selectedStyle)
```

This will result in the same behavior but more readable code.

app/src/main/kotlin/net/primal/android/core/compose/NostrUserText.kt
```diff
@@ -57,12 +58,18 @@ fun NostrUserText(
     maxLines: Int = 1,
     internetIdentifierBadgeSize: Dp = 14.dp,
     internetIdentifierBadgeAlign: PlaceholderVerticalAlign = PlaceholderVerticalAlign.Center,
-    customBadgeStyle: LegendaryStyle? = null,
+    authorLegendaryCustomization: LegendaryCustomization? = null,
```
markocic, line 61: Nitpick: I think `legendaryCustomization` is a better name here since we didn't use `author` prefix on other user-related data.

markocic (changes requested):
Very good! One more thing that should be changed along with comments is that every usage of `NostrUserText` should be given `legendaryCustomization` if there is also an `UniversalAvatarThumbnail` with `legendaryCustomization`. Examples of this can be seen in `MutedUserListItem.kt:129` and `NotificationListItem.kt:406`. Might be the only two places, if I haven't missed any.

markocic (approved):

AleksandarIlic (approved):
Good work!

PR #281, opened 2025-01-21
Implement zap links clickable in wallet tx details and fix duplicate link highlight

mehmedalijaK:
…links highlights


AleksandarIlic (changes requested):
Looks really good, just two minor comments.

AleksandarIlic (approved):

PR #284, opened 2025-01-22
Handle deep links for notes

markocic:


app/src/main/kotlin/net/primal/android/navigation/deeplinking/ext/DeepLinkingHandler.kt
```diff
@@ -1,66 +1,26 @@
...
+private val PRIMAL_NOTE_REGEX = Regex("https://.*primal.net/e/")
 
 private const val NOSTR_WALLET_CONNECT_SCHEMA = "nostr+walletconnect://"
 private const val NOSTR_WALLET_CONNECT_ALT_SCHEMA = "nostrwalletconnect://"
 
-fun String?.handleDeeplink(): DeepLink? {
-    if (this == null) return null
+fun String.parseDeepLinkOrNull(): DeepLink? =
+    when {
+        PRIMAL_NOTE_REGEX.containsMatchIn(this) ->
```
AleksandarIlic, line 15: `primal.net/e/` can be with `note1`, `noteHex` and `nevent1`. We should support all of them in the future, but for now we should ignore them.

Maybe yous should use `primal.net/e/note1` here and in `AndroidManfiest` to be on the safe side.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #287, opened 2025-01-27
Implement Wallet NWC Settings 

mehmedalijaK:


app/src/main/kotlin/net/primal/android/settings/wallet/WalletSettingsScreen.kt
```diff
@@ -160,6 +176,262 @@ fun WalletSettingsScreen(
...
+        Text(
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(horizontal = 2.dp)
+                .clickable(
+                    interactionSource = remember { MutableInteractionSource() },
+                    indication = null,
+                    onClick = createNewWalletConnection,
+                ),
+            text = buildAnnotatedString {
```
AleksandarIlic, line 382: Please use `TextButton` instead of clickable annotated string in the text.

We should render `Text` with the description and then below the `TextButton`.

And make sure that the font size is matching the design size. Right now is smaller than in design.

app/src/main/kotlin/net/primal/android/settings/wallet/WalletSettingsViewModel.kt
```diff
@@ -61,6 +80,14 @@ class WalletSettingsViewModel @Inject constructor(
                     is WalletSettingsContract.UiEvent.UpdateMinTransactionAmount -> {
                         updateSpamThresholdAmount(amountInSats = it.amountInSats)
                     }
+                    is WalletSettingsContract.UiEvent.RevokeConnection -> {
+                        val nwcPubKey = it.nwcPubkey
+                        val updatedConnections = state.value.connectedApps.filterNot { it.nwcPubkey == nwcPubKey }
+
+                        setState { copy(connectedApps = updatedConnections) }
+
+                        nwcWalletRepository.revokeConnection(activeAccountStore.activeUserId(), nwcPubKey)
```
AleksandarIlic, line 89: Extract this entire block in new coroutine.

And add try-catch to `revokeConnection` for `WssException`

app/src/main/kotlin/net/primal/android/settings/wallet/WalletSettingsViewModel.kt
```diff
@@ -133,4 +160,17 @@ class WalletSettingsViewModel @Inject constructor(
...
+
+    private fun PrimalNwcConnectionInfo.mapAsConnectedAppUi(): ConnectedAppUi {
+        return ConnectedAppUi(
+            nwcPubkey = nwcPubkey,
+            appName = appName,
+            dailyBudget =
+            if (dailyBudget?.isNotBlank() == true) {
+                dailyBudget.toSats().toLong().let { "%,d".format(it) }
+            } else {
+                "no limit"
```
AleksandarIlic, line 172: No strings allowed in ViewModels. In the model you can keep the sats which can be `nullable` to indicate no limit value. Then in the UI you should do the formatting and handling the `null` meaning.

app/src/main/res/values/strings.xml
```diff
@@ -690,12 +690,24 @@
     <string name="settings_wallet_max_wallet_balance">Max wallet balance</string>
     <string name="settings_wallet_max_wallet_balance_hint">Primal Wallet is a transactional wallet, designed for handling small amounts. We recommend self custody for larger amounts.</string>
     <string name="settings_wallet_hide_transactions_below">Hide transactions below</string>
+    <string name="settings_wallet_header_app">App</string>
+    <string name="settings_wallet_header_daily_budget">Daily Budget</string>
+    <string name="settings_wallet_header_revoke">Revoke</string>
+    <string name="settings_wallet_connected_apps">CONNECTED APPS</string>
```
AleksandarIlic, line 696: Write in `strings.xml` in regular caps. And then in Compose you can use `.toUpperCase()` (or some similar fun) to change the letters.

app/src/main/kotlin/net/primal/android/settings/wallet/WalletSettingsViewModel.kt
```diff
@@ -133,4 +160,17 @@ class WalletSettingsViewModel @Inject constructor(
                 walletRepository.deleteAllTransactions()
```
AleksandarIlic, line 160: We used to have `withContext(dispatcherProvider.io())` in ViewModels earlier, but then we decided to change the approach and make sure that repositories should always be safe to call from any thread. This means that ViewModel should not worry about thread when calling repository suspend fun.  


Let's keep this comment until everything is resolved and see if we can clean this viewModel a bit.

app/src/main/kotlin/net/primal/android/settings/wallet/WalletSettingsViewModel.kt
```diff
@@ -50,21 +52,59 @@ class WalletSettingsViewModel @Inject constructor(
...
+                    is UiEvent.UpdateMinTransactionAmount -> {
                         updateSpamThresholdAmount(amountInSats = it.amountInSats)
                     }
+
+                    is UiEvent.RevokeConnection -> {
+                        revokeNwcConnection(nwcPubkey = it.nwcPubkey)
+                    }
+
+                    UiEvent.RetryConnectionsFetch -> fetchWalletConnections()
+                    UiEvent.RequestFetchWalletConnections -> fetchWalletConnections()
```
AleksandarIlic, line 90: Two events doing the same job. Keep one and use it on both places.

app/src/main/kotlin/net/primal/android/settings/wallet/connection/NwcDailyBudgetBottomSheet.kt
```diff
@@ -0,0 +1,147 @@
...
+fun NwcDailyBudgetBottomSheet(
+    initialDailyBudget: String?,
+    onDismissRequest: () -> Unit,
+    onBudgetSelected: (String?) -> Unit,
+    sheetState: SheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true),
+    budgetOptions: List<String>,
+) {
+    val scope = rememberCoroutineScope()
+
+    val initialSelection = initialDailyBudget?.takeIf { it != "no limit" }
```
AleksandarIlic, line 49: This kind of checks are never good. When we need to change the logic here, we would have to go all over the code and find all occurrences of "no limit" and change it. It's not very flexible solution. Please do the business logic like this.

app/src/main/kotlin/net/primal/android/settings/wallet/connection/NwcNewWalletConnectionContract.kt
```diff
@@ -0,0 +1,23 @@
+package net.primal.android.settings.wallet.connection
+
+interface NwcNewWalletConnectionContract {
+    companion object {
+        val budgetOptions = listOf("1000", "10000", "100000", "1000000", "no limit")
```
AleksandarIlic, line 5: Keep all budget options as `Long?` (please use thousand indicator, e.g. 10_000) and remove "no limit" from budget options so we don't have to compare against that exact `String`.

`null` can represent no budget / unlimited budget option. This is actually how backend also works and we would be aligned with the same business logic here.

And update the `ViewModel` and events to work with `Long?` as input budget instead of `String` and we should be good.

AleksandarIlic (changes requested):
Just few more changes needed before we can merge it.

AleksandarIlic (approved):
This now looks great!

PR #289, opened 2025-01-30
Implement Link Primal Wallet Screen

mehmedalijaK:


app/src/main/res/values/strings.xml
```diff
@@ -727,7 +727,9 @@
     <string name="settings_wallet_new_nwc_connection_copy_nwc_string_button">Copy NWC String</string>
     <string name="settings_wallet_new_nwc_connection_done_button">Done</string>
     <string name="settings_wallet_new_nwc_connection_connection_created_text">Connection created</string>
-
+    <string name="settings_wallet_link_primal_wallet_title">Link Primal Wallet</string>
+    <string name="settings_wallet_link_external_app_request">An external app “%1$s” is requesting access to your Primal Wallet. If you wish to allow this, click “Give Wallet Access” below.</string>
+    <string name="settings_wallet_link_give_wallet_access">Give Wallet Access</string>
```
AleksandarIlic, line 732: nitpick: `_button` should be at the end.

app/src/main/kotlin/net/primal/android/settings/wallet/connection/NwcNewWalletConnectionScreen.kt
```diff
@@ -226,7 +233,7 @@ private fun WalletConnectionEditor(
                 modifier = Modifier.padding(horizontal = 32.dp),
                 text = stringResource(id = R.string.settings_wallet_new_nwc_connection_app_name_input_header),
                 color = AppTheme.colorScheme.onPrimary,
-                style = AppTheme.typography.bodyMedium.copy(fontSize = 16.sp),
+                style = AppTheme.typography.bodyLarge,
```
AleksandarIlic, line 236: Sometimes the font size from design is not matched to our internal typography. In this case we pick the nearest match, copy it and then modify it. This is why was it implemented like this.

`bodyLarge` is actually `18.sp`, while `bodyMedium` is `15.sp`.

You can see the entire typography at `net.primal.android.theme.AppTypography`.

I think in the design is `16.sp`. Please check and revert this is line if it's really `16.sp`.

app/src/main/kotlin/net/primal/android/settings/wallet/connection/NwcNewWalletConnectionScreen.kt
```diff
@@ -160,7 +156,7 @@ private fun NwcNewWalletConnectionScreen(
 }
 
 @Composable
-private fun NewWalletConnectionFooter(
+fun NewWalletConnectionFooter(
```
AleksandarIlic, line 159: Better to move this into it's own file.

app/src/main/kotlin/net/primal/android/settings/wallet/connection/NwcNewWalletConnectionScreen.kt
```diff
@@ -236,60 +243,18 @@ private fun WalletConnectionEditor(
...
-            }
-
             Spacer(modifier = Modifier.height(21.dp))
 
             Text(
                 modifier = Modifier.padding(horizontal = 21.dp),
                 text = stringResource(id = R.string.settings_wallet_new_nwc_connection_hint),
                 color = AppTheme.extraColorScheme.onSurfaceVariantAlt4,
-                style = AppTheme.typography.bodyMedium.copy(fontSize = 16.sp),
+                style = AppTheme.typography.bodyLarge,
```
AleksandarIlic, line 257: The same comment as for above. Please double check if it's really `18.sp` like in this typography.

app/src/main/kotlin/net/primal/android/settings/wallet/link/LinkPrimalWalletContract.kt
```diff
@@ -0,0 +1,26 @@
+package net.primal.android.settings.wallet.link
+
+interface LinkPrimalWalletContract {
+    companion object {
+        val budgetOptions: List<Long?> = listOf(1000, 10_000, 100_000, 1_000_000, null)
```
AleksandarIlic, line 5: Can we use the same const for both screens? Maybe put in new file and remove it from both contracts.

And in this case write `1000` as `1_000`

app/src/main/kotlin/net/primal/android/navigation/SettingsNavigation.kt
```diff
@@ -159,7 +188,14 @@ private fun NavGraphBuilder.wallet(
         onClose = { navController.navigateUp() },
         onEditProfileClick = { navController.navigateToProfileEditor() },
         onOtherConnectClick = { navController.navigateToWalletScanNwcUrl() },
-        onCreateNewWalletConnection = { navController.navigateToNewWalletConnection() },
+        onCreateNewWalletConnection = {
+//            navController.navigateToNewWalletConnection()
+            navController.navigateToLinkPrimalWallet(
```
AleksandarIlic, line 193: Did you tried calling it with only `callback` param? Does it work?

app/src/main/kotlin/net/primal/android/navigation/NavigationArguments.kt
```diff
@@ -77,6 +77,16 @@ const val NOTE_EDITOR_ARGS = "preFillContent"
...
 
+const val APP_NAME = "appName"
+inline val SavedStateHandle.appName: String? get() = get(APP_NAME)
+
+const val APP_ICON = "appIcon"
+inline val SavedStateHandle.appIcon: String? get() = get(APP_ICON)
+
+const val CALLBACK = "callback"
+inline val SavedStateHandle.callback: String
+    get() = get(CALLBACK) ?: throw IllegalArgumentException("Missing required $CALLBACK argument.")
```
AleksandarIlic, line 88: Can you please add `NWC` prefix to all `const val`s you added here?

app/src/main/kotlin/net/primal/android/core/compose/DailyBudgetPicker.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import net.primal.android.R
+import net.primal.android.theme.AppTheme
+
+@Composable
+fun DailyBudgetPicker(
+    modifier: Modifier = Modifier,
+    dailyBudget: Long?,
+    onChangeDailyBudgetBottomSheetVisibility: (bottomSheetVisible: Boolean) -> Unit,
```
AleksandarIlic, line 27: There is a rule for arguments order in `@Composable` functions:

Mandatory arguments go first.
Then optional arguments go second (the one with default value).

If `modifier` is there, then it's first in either mandatory or optional group.

app/src/main/kotlin/net/primal/android/settings/wallet/link/LinkPrimalWalletScreen.kt
```diff
@@ -0,0 +1,265 @@
...
+        topBar = {
+            PrimalTopAppBar(
+                title = stringResource(id = R.string.settings_wallet_link_primal_wallet_title),
+                navigationIcon = PrimalIcons.ArrowBack,
+                navigationIconContentDescription = stringResource(id = R.string.accessibility_back_button),
+                onNavigationIconClick = onClose,
+            )
+        },
+        content = { paddingValues ->
+            AnimatedContent(targetState = state.secret) { secret ->
```
AleksandarIlic, line 91: No need for `AnimatedContent`. We are always going to show this screen.

And no need to have `secret` in the Contract since we will be passing it back through SideEffect which you implemented nicely, so please remove `secret` from contract. Unless I missed something here.

app/src/main/kotlin/net/primal/android/settings/wallet/link/LinkPrimalWalletScreen.kt
```diff
@@ -0,0 +1,265 @@
...
+                color = AppTheme.extraColorScheme.onSurfaceVariantAlt4,
+                style = AppTheme.typography.bodyLarge,
+                textAlign = TextAlign.Center,
+            )
+        }
+    }
+}
+
+@Composable
+private fun WalletConnectionEditorHeader(
```
AleksandarIlic, line 190: Since we are already moving out some shared components, does it makes sense this to be a shared component for both screens?

I originally thought that won't be necessary, but now I'm no longer sure.

app/src/main/kotlin/net/primal/android/settings/wallet/link/LinkPrimalWalletViewModel.kt
```diff
@@ -0,0 +1,107 @@
...
+import timber.log.Timber
+
+@HiltViewModel
+class LinkPrimalWalletViewModel @Inject constructor(
+    savedStateHandle: SavedStateHandle,
+    private val activeAccountStore: ActiveAccountStore,
+    private val nwcWalletRepository: NwcWalletRepository,
+) : ViewModel() {
+
+    private val appName: String = savedStateHandle.appName ?: "External app"
```
AleksandarIlic, line 35: Please put `External app` as `const` in `companion object` of this ViewModel.

AleksandarIlic (changes requested):
Most of the comments are nit picks. This is really great implementation. Great job!

app/src/main/kotlin/net/primal/android/core/compose/DailyBudgetPicker.kt
AleksandarIlic: @mehmedalijaK This does not look resolved. Optional modifier should be first after listing first mandatory arguments. In this case, it should be at the end because other two are mandatory (there is no default param).

app/src/main/kotlin/net/primal/android/settings/wallet/link/LinkPrimalWalletViewModel.kt
```diff
@@ -0,0 +1,107 @@
...
+
+                val formattedDailyBudgetBtc = dailyBudgetBtc?.let {
+                    BigDecimal(it.toString())
+                        .stripTrailingZeros()
+                        .toPlainString()
+                }
+
+                val response = nwcWalletRepository.createNewWalletConnection(
+                    userId = activeAccountStore.activeUserId(),
+                    appName = state.value.appName ?: "External app",
```
AleksandarIlic, line 90: You should use const here you defined in contract I guess.

app/src/main/kotlin/net/primal/android/core/compose/WalletConnectionEditorHeader.kt
```diff
@@ -0,0 +1,106 @@
+package net.primal.android.core.compose
```
AleksandarIlic, line 1: Can you please move it in `settings.wallet` package? It won't be used outside that feature.

app/src/main/kotlin/net/primal/android/core/compose/WalletConnectionFooter.kt
```diff
@@ -0,0 +1,62 @@
+package net.primal.android.core.compose
```
AleksandarIlic, line 1: Can you please move it in `settings.wallet` package? It won't be used outside that feature.

AleksandarIlic (changes requested):

AleksandarIlic (approved):
Good job!

PR #312, opened 2025-02-10
Migrate database models to support multi-account

markocic:


app/src/main/kotlin/net/primal/android/feeds/db/Feed.kt
```diff
@@ -1,12 +1,15 @@
...
+import androidx.room.Index
 import androidx.room.PrimaryKey
 import net.primal.android.feeds.domain.FeedSpecKind
 
-@Entity
+@Entity(indices = [Index(value = ["ownerId", "spec"], unique = true)])
 data class Feed(
-    @PrimaryKey
+    @PrimaryKey(autoGenerate = true)
+    val order: Int = 0,
```
AleksandarIlic, line 11: Maybe we should use some other name like `position` instead of `order` which is reserved keyword in SQL.

app/src/main/kotlin/net/primal/android/profile/db/ProfileInteractionDao.kt
```diff
@@ -6,38 +6,24 @@ import androidx.room.OnConflictStrategy
...
-    @Transaction
-    fun insertOrUpdate(interaction: ProfileInteraction) {
-        val existingProfile = getInteractionByProfileId(interaction.profileId)
-        if (existingProfile != null) {
-            update(interaction)
-        } else {
-            insert(interaction)
-        }
-    }
+    @Upsert
```
AleksandarIlic, line 15: `@Upsert` was causing problems with database encryption in the past. This was the reason we are not using in the app.

AleksandarIlic: I'm not sure if this is still valid though with newest version. This was the bug:
https://github.com/sqlcipher/sqlcipher-android/issues/23

AleksandarIlic (approved):

PR #319, opened 2025-02-12
Enforce minimal contribution amount

mehmedalijaK:


app/src/main/kotlin/net/primal/android/premium/legend/contribute/LegendContributeContract.kt
```diff
@@ -5,6 +5,10 @@ import net.primal.android.wallet.domain.CurrencyMode
 
 class LegendContributeContract {
```
AleksandarIlic, line 6: This should actually be interface. Contracts should always be `interface`s. Please fix it.

AleksandarIlic (approved):
Please fix the comment in your last PR for this feature.

PR #323, opened 2025-02-14
Implement multi-account feature

markocic:
Closes: https://github.com/PrimalHQ/primal-android-app/issues/2, Closes https://github.com/PrimalHQ/primal-android-app/issues/249


app/schemas/net.primal.android.db.PrimalDatabase/56.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 56,
-    "identityHash": "a279261907a560873ca989dd79346c3a",
+    "identityHash": "2c54cd4e6f2437637cee188d1ba11935",
```
AleksandarIlic, line 5: Please increase the database version and keep `56` as it is.

app/src/main/kotlin/net/primal/android/crypto/ConversionUtils.kt
```diff
@@ -4,6 +4,8 @@ import org.spongycastle.util.encoders.DecoderException
 import org.spongycastle.util.encoders.Hex
 import timber.log.Timber
 
+fun String.assureValidNsec() = if (startsWith("nsec")) this else this.hexToNsecHrp()
```
AleksandarIlic, line 7: Does this need to be public?

app/src/main/kotlin/net/primal/android/drawer/multiaccount/AccountSwitcher.kt
```diff
@@ -0,0 +1,116 @@
...
+@Composable
+private fun AccountSwitcher(
+    modifier: Modifier = Modifier,
+    eventPublisher: (AccountSwitcherContract.UiEvent) -> Unit,
+    state: AccountSwitcherContract.UiState,
+    callbacks: AccountSwitcherCallbacks,
+    onLogoutClick: (String) -> Unit,
+) {
+    var accountsBottomSheetVisibility by remember { mutableStateOf(false) }
+    val bottomSheetIcon = remember(state.userAccounts) {
```
AleksandarIlic, line 67: `remember` is expensive operation. We should not use it for simple `if` check. Just assigned it to `val` or use the `if` directly.

app/src/main/kotlin/net/primal/android/explore/repository/ExploreRepository.kt
```diff
@@ -35,12 +35,14 @@ import net.primal.android.nostr.ext.takeContentAsPrimalUserFollowStats
...
 import net.primal.android.nostr.ext.takeContentAsPrimalUserScoresOrNull
 import net.primal.android.profile.db.ProfileStats
+import net.primal.android.user.accounts.active.ActiveAccountStore
 import net.primal.android.wallet.utils.CurrencyConversionUtils.toSats
 
 class ExploreRepository @Inject constructor(
     private val dispatchers: CoroutineDispatcherProvider,
     private val exploreApi: ExploreApi,
     private val database: PrimalDatabase,
+    private val activeAccountStore: ActiveAccountStore,
```
AleksandarIlic, line 45: I think we should not use `ActiveAccountStore` in repositories. From repository PoV it doesn't matter which user is active. Instead just pass the `userId` as an argument to the `fun`.

app/src/main/kotlin/net/primal/android/explore/repository/ExploreRepository.kt
```diff
@@ -207,7 +209,8 @@ class ExploreRepository @Inject constructor(
         }
 
     fun observeRecentUsers(): Flow<List<UserProfileSearchItem>> {
-        return database.profileInteractions().observeRecentProfiles()
+        return database.profileInteractions()
+            .observeRecentProfilesByOwnerId(ownerId = activeAccountStore.activeUserId())
```
AleksandarIlic, line 213: Pass `userId` as an argument instead.

app/src/main/kotlin/net/primal/android/explore/repository/ExploreRepository.kt
AleksandarIlic: If this is used in some other repositories, then we should fix that as well.

app/src/main/kotlin/net/primal/android/messages/conversation/MessageConversationListViewModel.kt
```diff
@@ -44,7 +44,7 @@ class MessageConversationListViewModel @Inject constructor(
         value = UiState(
             activeRelation = ConversationRelation.Follows,
             conversations = messageRepository
-                .newestConversations(ConversationRelation.Follows)
+                .newestConversations(userId = activeUserId, relation = ConversationRelation.Follows)
```
AleksandarIlic, line 47: `private val activeUserId = activeAccountStore.activeUserId()`

Let's remove the line above (line 41) and use `activeAccountStore.activeUserId()` directly in the code as we do in other viewModels. I don't know why I cached this value in the first place.

app/src/main/kotlin/net/primal/android/user/repository/UserRepository.kt
```diff
@@ -35,10 +38,20 @@ class UserRepository @Inject constructor(
...
 ) {
+    suspend fun setActiveAccount(userId: String) =
+        withContext(dispatchers.io()) {
+            accountsStore.getAndUpdateAccount(userId = userId) { copy(lastAccessedAt = Instant.now().epochSecond) }
+            activeAccountStore.setActiveUserId(pubkey = userId)
+        }
+
+    fun observeUserAccounts() = accountsStore.userAccounts
+
+    fun observeActiveAccount() = activeAccountStore.activeUserAccount.distinctUntilChanged()
```
AleksandarIlic, line 54: Not sure if we really need `observeActiveAccount` and `observeUserAccounts` here. 

We could always get it directly from `ActiveAccountStore` whenever we need it.

AleksandarIlic (changes requested):
Great work! Just a couple of minor comments + functionality fixes we already talked about.

app/src/main/kotlin/net/primal/android/crypto/ConversionUtils.kt
markocic: We are using it in `LoginHandler.kt:33` as well.

AleksandarIlic (approved):

PR #327, opened 2025-02-19
Implement primal.net/legends clicks for legends screen

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteContent.kt
```diff
@@ -219,6 +219,7 @@ fun NoteContent(
                         onPostClick = noteCallbacks.onNoteClick,
                         onHashtagClick = noteCallbacks.onHashtagClick,
                         onArticleClick = noteCallbacks.onArticleClick,
+                        onPrimalLegendsLeaderboardClick = noteCallbacks.onPrimalLegendsLeaderboardClick,
```
AleksandarIlic, line 222: `PrimalLegend` is not an annotation, but a special handling of URL annotation. 

Instead of passing a custom click listener and doing the logic in `handleAnnotationClick`, you can take that logic and do the check here in `onUrlClick = { }`.

AleksandarIlic (changes requested):
Good work. Just minor logic move change requested.

AleksandarIlic (approved):

PR #330, opened 2025-02-20
Implement nevent and nprofile in the note editor for mentioned notes and users

markocic:


app/src/main/kotlin/net/primal/android/user/api/UsersApi.kt
```diff
@@ -27,6 +28,8 @@ interface UsersApi {
 
     suspend fun getUserRelays(userId: String): UserRelaysResponse
 
+    suspend fun getUserRelays(userIds: List<String>): UsersRelaysResponse
```
AleksandarIlic, line 31: Can we merge `getUserRelays(userId: String)` with this one?

AleksandarIlic (changes requested):
Max 2 relays in nprofiles and nevents.

PR #347, opened 2025-02-28
Refactor splash screen and reimplement existing deep links

markocic:


app/src/main/kotlin/net/primal/android/settings/wallet/nwc/primal/link/LinkPrimalWalletViewModel.kt
```diff
@@ -28,22 +27,26 @@ import net.primal.android.wallet.repository.NwcWalletRepository
...
-    private val appName: String = savedStateHandle.appName?.takeIf { it.isNotEmpty() } ?: DEFAULT_APP_NAME
-    private val appIcon: String? = savedStateHandle.appIcon
-    private val callback: String = savedStateHandle.callback
+    @AssistedFactory
+    interface Factory {
+        fun create(nwcPrimalUrl: String?): LinkPrimalWalletViewModel
+    }
+
+    private val primalNwcWallet = nwcPrimalUrl?.parseAsPrimalWalletNwc()
+        ?: error("could not parse nwc primal wallet uri.")
```
AleksandarIlic, line 43: I'm not sure if we should to `error` here since it would crash the app if users uses invalid url.

If we can not prevent entering here with invalid url then we should handle the invalid url case as well.

app/src/main/kotlin/net/primal/android/thread/notes/ThreadViewModel.kt
```diff
@@ -139,4 +141,11 @@ class ThreadViewModel @Inject constructor(
                 Timber.w(error)
             }
         }
+
+    private fun String.resolveNoteIdOrThrow(): String =
+        when {
+            this.startsWith("note1") -> runCatching { bech32ToHexOrThrow() }.getOrNull()
+            this.startsWith("nevent1") -> Nip19TLV.parseUriAsNeventOrNull(this)?.eventId
+            else -> this
+        } ?: error("couldn't resolve noteId")
```
AleksandarIlic, line 150: We should not throw errors instead we should gracefully handle this situation. Imagine someone opening thread screen with invalid link and the app crashes.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

AleksandarIlic (approved):

PR #350, opened 2025-03-03
Add more context menu options to media gallery screen

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/utils/ClipboardUtils.kt
```diff
@@ -13,3 +24,45 @@ fun copyText(
...
+            } catch (e: MalformedURLException) {
+                throw IllegalArgumentException("Invalid URL: $imageUrl", e)
+            }
+
+            val connection = url.openConnection()
+            connection.connect()
+
+            val inputStream = connection.getInputStream()
+            val file = File(
+                context.getExternalFilesDir(Environment.DIRECTORY_PICTURES),
```
AleksandarIlic, line 42: We should put in a cache folder instead. This will make it visible in the gallery app when you copy something.

app/src/main/kotlin/net/primal/android/core/utils/ClipboardUtils.kt
```diff
@@ -13,3 +24,45 @@ fun copyText(
...
+                "copied_image.png",
```
AleksandarIlic, line 43: Maybe we should not use an extension here since you don't know if it's going to be a png or something else.

app/src/main/kotlin/net/primal/android/core/utils/ClipboardUtils.kt
```diff
@@ -13,3 +24,45 @@ fun copyText(
     val clip = ClipData.newPlainText(label, text)
     clipboard.setPrimaryClip(clip)
 }
+
+suspend fun copyImageToClipboard(context: Context, imageUrl: String) {
```
AleksandarIlic, line 28: I'm not sure if this is the best way to handle this situation. We already have a loaded image through Coil.

Maybe we can take the already rendered `Bitmap` and store it in a cache folder of the app. Then use that file to expose Uri with `FileProvider.getUriForFile` as you did below.

app/src/main/kotlin/net/primal/android/attachments/gallery/MediaGalleryScreen.kt
```diff
@@ -162,6 +167,16 @@ fun MediaGalleryScreen(
                         onSaveClick = {
                             currentImage()?.let { eventPublisher(MediaGalleryContract.UiEvent.SaveMedia(it)) }
                         },
+                        onMediaUrlCopyClick = {
+                            currentImage()?.url?.let { copyText(context = context, text = it) }
```
AleksandarIlic, line 171: We should probably copy the original url of the image and not the primal cdn. 

To which url is `currentImage()?.url` pointing to? Is it original or cdn link?

AleksandarIlic (changes requested):
Looks really good! Just few concerns regarding how we copy the image to clipboard.

AleksandarIlic (changes requested):
Test if MediaItemScreen was not broken with my changes.

We also need a working solution for currently visible `Bitmap` in horizontal pager of MediaGallery.

AleksandarIlic:
Our Horizontal Pager setup can hold up to 3 pages at the same time (previous, current and previous) and I'm not sure if this solution will work correctly in all cases.

Let's just remove copy bitmap from gallery from this PR and merge everything else.

AleksandarIlic (approved):

PR #351, opened 2025-03-03
Handle loading and error states in profileId resolution process

markocic:


app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsContract.kt
```diff
@@ -27,11 +27,18 @@ interface ProfileDetailsContract {
             ProfileFeedSpec.AuthoredArticles,
             ProfileFeedSpec.AuthoredMedia,
         ),
+        val isResolvingProfileId: Boolean = true,
```
AleksandarIlic, line 30: This could be just below `profileId` since they are coupled. It makes it easier to read the code later.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsViewModel.kt
```diff
@@ -85,18 +75,44 @@ class ProfileDetailsViewModel @Inject constructor(
...
-        markProfileInteraction()
+        initializeProfileId()
     }
 
-    private fun markProfileInteraction() {
+    private fun initializeProfileId() =
+        viewModelScope.launch {
+            try {
+                setState { copy(isResolvingProfileId = true, resolutionError = null) }
+                val profileId = withContext(dispatcherProvider.io()) { resolveProfileId() }!!
```
AleksandarIlic, line 86: Let's not use `!!`.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsViewModel.kt
```diff
@@ -85,18 +75,44 @@ class ProfileDetailsViewModel @Inject constructor(
...
+            } catch (error: UnknownHostException) {
+                setState { copy(resolutionError = UiState.ProfileResolutionError.ErrorResolvingPrimalName(error)) }
+            } catch (_: RuntimeException) {
+                setState { copy(resolutionError = UiState.ProfileResolutionError.InvalidProfileId) }
+            } finally {
+                setState { copy(isResolvingProfileId = false) }
+            }
+        }
+
+    private fun initializeProfileData(profileId: String, isActiveUser: Boolean) {
```
AleksandarIlic, line 104: The name is a bit misleading. We are doing a bunch of things here and some are not even related to the profile data. We'll need better naming for this function and for `initializeProfileId` as well. I don't have better suggestion, yet. If you have any alternatives, now is the time.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsViewModel.kt
AleksandarIlic: And I'm thinking of using the `resolveProfileId` inline here. It's just two more lines. Or at least `resolveProfileId` could be an extension function on `SavedStateHandle`.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsViewModel.kt
```diff
@@ -85,18 +75,44 @@ class ProfileDetailsViewModel @Inject constructor(
...
+
+                if (profileId.isValidHex()) {
+                    setState { copy(profileId = profileId, isActiveUser = isActiveUser) }
+                    initializeProfileData(profileId = profileId, isActiveUser = isActiveUser)
+                } else {
+                    setState { copy(resolutionError = UiState.ProfileResolutionError.InvalidProfileId) }
+                }
+            } catch (error: UnknownHostException) {
+                setState { copy(resolutionError = UiState.ProfileResolutionError.ErrorResolvingPrimalName(error)) }
+            } catch (_: RuntimeException) {
```
AleksandarIlic, line 97: Maybe we just wrap this around `runCatching`.  

And I'm not really sure if we should go this much in depth of showing what the actual problem is with profile resolution. In case of an error we know that we could not load the profile and that should be enough since it's now a prime time feature. This is going to happen only in edge-cases, very rarely.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsScreen.kt
```diff
@@ -110,7 +112,12 @@ fun ProfileDetailsScreen(
 
     DisposableLifecycleObserverEffect(viewModel) {
         when (it) {
-            Lifecycle.Event.ON_START -> viewModel.setEvent(ProfileDetailsContract.UiEvent.RequestProfileUpdate)
+            Lifecycle.Event.ON_START -> {
+                uiState.value.profileId?.let {
```
AleksandarIlic, line 116: UiState is already in the viewModel. Why doing this? We could just read it directly there from the state?

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsScreen.kt
```diff
@@ -374,174 +381,230 @@ fun ProfileDetailsScreen(
...
-                                        message = context.getString(
-                                            R.string.wallet_send_payment_error_nostr_user_without_lightning_address,
-                                            state.profileDetails?.authorDisplayName
-                                                ?: context.getString(R.string.wallet_send_payment_this_user_chunk),
-                                        ),
-                                        duration = SnackbarDuration.Short,
+                if (state.isResolvingProfileId) {
+                    PrimalLoadingSpinner()
+                } else if (state.resolutionError != null) {
+                    when (state.resolutionError) {
```
AleksandarIlic, line 392: We could have an `@Composable` function which would render resolution error. It would reduce this to few lines by hiding the implementation details.

Also, already suggested, we should just have a one simple error for resolution and that's it. You already doing the same action.

Off-topic: If two different errors would be needed, the following error handling code should just be one `ListNoContent` and the `noContentText` would be resolved based on the error type (`when` would go there) since everything else is already the same.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsScreen.kt
```diff
@@ -374,174 +381,230 @@ fun ProfileDetailsScreen(
...
-                                state = pagerState,
-                            ) { pageIndex ->
-                                when {
-                                    state.isProfileMuted -> {
-                                        Box(
-                                            modifier = Modifier.fillMaxSize(),
+                        }
+                    }
+                } else {
+                    LazyColumn(
```
AleksandarIlic, line 418: Can we put everything from this `else` block into it's own `@Composable` function where `profileId` would mandatory field? What are the challenges?

This might reduce a bit the complexity of `ProfileDetailsScreen(state)` which is the most complex `@Composable fun` in the entire project I think/hope.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsContract.kt
```diff
@@ -27,11 +27,18 @@ interface ProfileDetailsContract {
...
+        val resolutionError: ProfileResolutionError? = null,
         val error: ProfileError? = null,
         val shouldApproveProfileAction: ProfileApproval? = null,
         val zapError: UiError? = null,
         val zappingState: ZappingState = ZappingState(),
     ) {
+        sealed class ProfileResolutionError {
```
AleksandarIlic, line 37: Maybe a simple `Boolean` is enough for this error. See other coments.

app/src/main/kotlin/net/primal/android/profile/details/ui/ProfileDetailsHeader.kt
```diff
@@ -73,7 +73,7 @@ fun ProfileDetailsHeader(
     ProfileHeaderDetails(
         state = state,
         onEditProfileClick = onEditProfileClick,
-        onMessageClick = { onMessageClick(state.profileId) },
+        onMessageClick = { state.profileId?.let { onMessageClick(state.profileId) } },
```
AleksandarIlic, line 76: I would put `profileId: String` as an argument of function and use that one to avoid lets everywhere. 

Or better, we could do `check(state.profileId != null)` or whatever is the syntax. To make it a contract that `ProfileDetailsHeader` can not be used if state is not having `profileId`.

AleksandarIlic (changes requested):
Good changes with room for improvement. I like the approach in general. Check the comments to see where we can simplify implementation and improve.

AleksandarIlic (approved):

PR #354, opened 2025-03-04
Improve gallery rendering for media attachments in notes

mehmedalijaK:


AleksandarIlic (approved):
Thanks! We should wait with merge until we ship current release.

PR #357, opened 2025-03-08
Fix new notes pill doesn't show default avatars

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedViewModel.kt
```diff
@@ -203,7 +203,9 @@ class NoteFeedViewModel @AssistedInject constructor(
 
         val newSyncStats = FeedPostsSyncStats(
             latestNoteIds = allNotes.map { it.id },
-            latestAvatarCdnImages = avatarCdnImagesAndLegendaryCustomizations.mapNotNull { it.first }.take(limit),
+            latestAvatarCdnImages = avatarCdnImagesAndLegendaryCustomizations
+                .map { it.first ?: CdnImage(sourceUrl = "") }
```
AleksandarIlic, line 207: I'm not sure if this empty `CdnImage` is the best solution. What would happen if we would allow nulls and then just pass `null` for those who does not have image?

If this is just complicating things, then we maybe could do with `CdnImage` something like we have with `UserAccount` - adding `EMPTY` to `companion object` of `CdnImage`.

PR #360, opened 2025-03-10
Wallet Receive: Tapping on the bitcoin or lightning address should copy that info/text

mehmedalijaK:


app/src/main/kotlin/net/primal/android/wallet/transactions/receive/ReceivePaymentScreen.kt
```diff
@@ -332,26 +334,33 @@ private fun ReceivePaymentViewer(
...
-                    ReceivePaymentTab.Lightning -> networkDetails.address
-                    ReceivePaymentTab.Bitcoin -> networkDetails.address.ellipsizeMiddle(size = 12)
-                },
-                contentFontSize = 20.sp,
-                maxLines = when (state.currentTab) {
-                    ReceivePaymentTab.Lightning -> 3
-                    ReceivePaymentTab.Bitcoin -> 1
-                },
-            )
+            Column(
```
AleksandarIlic, line 339: The column should be inside of `TwoLineText`. The **content** of `TwoLineText` should be in Column. And `TwoLineText` will then have `onClick` argument.

AleksandarIlic (changes requested):

PR #361, opened 2025-03-10
Wallet Receive: Tapping the Copy button on Lighting tab should copy `lnurl` value from QR code and not lud16 (email like) address

mehmedalijaK:


AleksandarIlic (approved):

app/src/main/kotlin/net/primal/android/wallet/transactions/receive/model/NetworkDetails.kt
```diff
@@ -24,5 +24,5 @@ data class NetworkDetails(
             }
         }
 
-    val copyValue: String? get() = invoice ?: address
+    val copyValue: String? get() = invoice ?: qrCodeValue
```
AleksandarIlic, line 27: We actually need copy value to be without prefix `lightning:` and `bitcoin:`. I guess it's better to reconstruct it.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #366, opened 2025-03-13
Feature/login via npub

markocic:


app/src/main/kotlin/net/primal/android/nostr/ext/DirectMessageEvents.kt
```diff
@@ -35,18 +35,18 @@ fun PrimalEvent.asMessageConversationsSummary(): ConversationsSummary {
...
+fun NostrEvent.mapAsMessageDataPO(userId: String, nsec: String?): DirectMessageData? {
     val senderId = this.pubKey
     val receiverId = this.tags.findFirstProfileId() ?: return null
     val participantId = if (senderId != userId) senderId else receiverId
 
     val decryptedMessage = runCatching {
         CryptoUtils.decrypt(
             message = this.content,
-            privateKey = nsec.bechToBytesOrThrow(hrp = "nsec"),
+            privateKey = nsec?.bechToBytesOrThrow(hrp = "nsec")!!, /* TODO(marko): what to do here? */
```
AleksandarIlic, line 49: I would throw the missing key exception if it's `null`. We have `getOrElse` in place so we are good. With this approach we can later check the exception in the block below and customize the content if we want.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
```diff
@@ -35,7 +34,12 @@ class NostrNotary @Inject constructor(
     private fun findNsecOrThrow(pubkey: String): String {
         return try {
             val npub = Hex.decode(pubkey).toNpub()
-            credentialsStore.findOrThrow(npub = npub).nsec
+            val nsec = credentialsStore.findOrThrow(npub = npub).nsec
+            if (nsec == null) {
```
AleksandarIlic, line 38: Can you do `?: throw MissingPrivateKeyException()` here?

app/src/main/kotlin/net/primal/android/premium/manage/order/PremiumOrderHistoryViewModel.kt
```diff
@@ -99,6 +102,9 @@ class PremiumOrderHistoryViewModel @Inject constructor(
                 try {
                     premiumRepository.cancelSubscription(userId = userId, purchaseJson = purchase.playSubscriptionJson)
                     premiumRepository.fetchMembershipStatus(activeAccountStore.activeUserId())
+                } catch (error: MissingPrivateKeyException) {
+                    /* TODO(marko): not sure if we should do something here, I don't think the user can get here, silent fail might be good enough */
```
AleksandarIlic, line 106: User can't reach this point with npub login. Just ignore it.

app/src/main/kotlin/net/primal/android/premium/repository/PremiumRepository.kt
```diff
@@ -45,11 +47,17 @@ class PremiumRepository @Inject constructor(
...
-                }
-            }?.premiumMembership
+            try {
+                premiumApi.getPremiumMembershipStatus(userId = userId)?.let { response ->
+                    accountsStore.getAndUpdateAccount(userId = userId) {
+                        this.copy(premiumMembership = response.toPremiumMembership())
+                    }
+                }?.premiumMembership
+            } catch (error: MissingPrivateKeyException) {
+                /* TODO(marko): should we catch this here and fail silently, or handle this in viewModel? */
```
AleksandarIlic, line 57: All premium stuff can be safely ignored.

app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsViewModel.kt
```diff
@@ -374,6 +375,10 @@ class ProfileDetailsViewModel @Inject constructor(
                 Timber.w(error)
                 updateStateProfileAsUnfollowedAndClearApprovalFlag()
                 setErrorState(error = UiError.FailedToFollowUser(error))
+            } catch (error: MissingPrivateKeyException) {
+                Timber.w(error)
+                updateStateProfileAsUnfollowedAndClearApprovalFlag()
+                setErrorState(error = UiError.FailedToFollowUser(error))
```
AleksandarIlic, line 381: This is the approach I want where we have `UiError` in place. 

The only difference is that we want to show another exception which says something that user is logged in with npub.

This comment applies for every viewmodel where we have `UiError`

app/src/main/kotlin/net/primal/android/settings/keys/KeysSettingsViewModel.kt
```diff
@@ -35,7 +35,7 @@ class KeysSettingsViewModel @Inject constructor(
                     val credential = credentialsStore.findOrThrow(it.pubkey.hexToNpubHrp())
                     copy(
                         avatarCdnImage = it.avatarCdnImage,
-                        nsec = credential.nsec,
+                        nsec = credential.nsec ?: "", /* TODO(marko): what to do here? */
```
AleksandarIlic, line 38: Just make `nsec` nullable in `UiState`. And on the screen if nsec is missing, do not render the nsec section.

app/src/main/kotlin/net/primal/android/settings/repository/SettingsRepository.kt
```diff
@@ -114,12 +116,16 @@ class SettingsRepository @Inject constructor(
...
-        )
-    }
+    private suspend fun fetchAppSettings(userId: String): ContentAppSettings? =
+        try {
+            val response = settingsApi.getAppSettings(pubkey = userId)
+            NostrJson.decodeFromStringOrNull<ContentAppSettings>(
+                string = response.userSettings?.content ?: response.defaultSettings?.content,
+            )
+        } catch (error: MissingPrivateKeyException) {
+            Timber.w(error) /* TODO(marko): should we catch this here? */
```
AleksandarIlic, line 126: I think this is fine and easier.

Alternative is to propagate it and put `@Throws` on this and public functions.

app/src/main/kotlin/net/primal/android/settings/wallet/nwc/primal/create/CreateNewWalletConnectionViewModel.kt
```diff
@@ -85,6 +86,8 @@ class CreateNewWalletConnectionViewModel @Inject constructor(
                         creatingSecret = false,
                     )
                 }
+            } catch (error: MissingPrivateKeyException) {
+                Timber.w(error)
```
AleksandarIlic, line 90: You should probably call `setState { copy(creatingSecret = false)` here as well.

app/src/main/kotlin/net/primal/android/user/updater/UserDataUpdater.kt
```diff
@@ -42,6 +43,7 @@ class UserDataUpdater @AssistedInject constructor(
     }
 
     private suspend fun updateData() {
+        /* TODO(marko): maybe all of this should go in a try catch? */
```
AleksandarIlic, line 46: This is `private fun`. Maybe you can move the catches into  `updateUserDataWithDebounce`?

It looks like if one fails all others are failing also.

AleksandarIlic (changes requested):
I would put additional error in `UiError` and whenever we are using int screen state I would set the missing key exception to that error so it can be properly displayed on the screen.

In cases where we use the custom screen error, I would just reuse the most logical one.

And where we don't have error handler, I would just ignore it.

The 3 buttons on the wallet dashboard should be disabled if user is logged in with npub. Check this in the viewModel, put in the state and take into consideration when rendering buttons.

Also, we can merge this directly into `main`. The solution is really good and the app works great. This means that all TODO needs to be resolved. I think all the comments are addressing directly or indirectly all TODO comments.

PR #371, opened 2025-03-14
Fix crash NoSuchElementException in NoteEditorViewModel

mehmedalijaK:


AleksandarIlic:
@markocic Can you have quick look on this one-liner PR? You worked on this `concatenateReferencedEvents` few weeks ago. The fix looks legit to me, but wanted to double check with you also.

markocic (changes requested):
I think I figured out the flow that could cause this crash to happen:
1. User wants to quote a note.
2. Thread in which this note appears is not cached locally and the fetching is still not completed. (Which causes the `conversation` array to be empty)
3. User writes his quote and publishes it. When we try to construct a quote event we look in the `conversion` list for necessary information which leads to this exception.

This can be reproduced on a slow internet connection.

This fix does mitigate the crash but doesn't address the real issue. This would cause the published quote event to be invalid by excluding a note which we are quoting.

My opinion is that creating an invalid quote event or having a crash in production are not acceptable options here.

Suggested fix:
We can prevent user from posting until we gather all the necessary information by disabling `Post` button.
We can introduce a "fetching" state in this screen and change button to be disabled with text: `Loading...` to indicate to the user the reason we have disabled that button.

PR #374, opened 2025-03-19
Introduce multi-module architecture with `shared` module for shared logic between platforms

AleksandarIlic:


AleksandarIlic:
This PR was merged manually.

PR #376, opened 2025-03-22
Implement Login with Amber

markocic:
Actual signing of the events is still left to be implemented. This type of login will be treated as npub login for now. User will be prompted to only sign one event that is part of the existing login process.


app/src/main/kotlin/net/primal/android/auth/login/LauncherExts.kt
```diff
@@ -0,0 +1,36 @@
...
+fun SignerLauncher.launchGetPublicKey() {
+    val intent = Intent(Intent.ACTION_VIEW, URI_PREFIX.toUri())
+    intent.`package` = PACKAGE_NAME
+
+    intent.putExtra("type", SignerMethod.GET_PUBLIC_KEY.method)
+    launch(intent)
+}
+
+fun SignerLauncher.launchSignEvent(event: NostrUnsignedEvent) {
+    val intent = Intent(Intent.ACTION_VIEW, "$URI_PREFIX${CommonJson.encodeToString(event)}".toUri())
```
AleksandarIlic, line 29: We have now `event.encodeToJsonString()` and it should be preferred way to encoding to json string. It uses the CommonJson by default.

app/src/main/kotlin/net/primal/android/auth/login/LoginScreen.kt
```diff
@@ -57,29 +60,44 @@ import net.primal.android.core.compose.AppBarIcon
...
 @Composable
 fun LoginScreen(
     viewModel: LoginViewModel,
     onClose: () -> Unit,
     onLoginSuccess: () -> Unit,
 ) {
+    val signLauncher = rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.StartActivityForResult(),
+    ) { result ->
+        if (result.resultCode != Activity.RESULT_OK) return@rememberLauncherForActivityResult
```
AleksandarIlic, line 89: nitpick: I find it easier to read it if just put the code below within proper `if` condition.

app/src/main/kotlin/net/primal/android/auth/login/LoginScreen.kt
```diff
@@ -57,29 +60,44 @@ import net.primal.android.core.compose.AppBarIcon
...
+
+        val nostrEvent = CommonJson.decodeFromStringOrNull<NostrEvent>(result.data?.getStringExtra("event"))
```
AleksandarIlic, line 91: For CommonJson stuff, we also have `String.decodeFromJsonStringOrNull()` which should be preferred instead of using Json directly.

app/src/main/kotlin/net/primal/android/auth/login/LoginScreen.kt
```diff
@@ -104,6 +122,16 @@ fun LoginScreen(
     eventPublisher: (LoginContract.UiEvent) -> Unit,
     onClose: () -> Unit,
 ) {
+
+    val pubkeyLauncher = rememberLauncherForActivityResult(
```
AleksandarIlic, line 126: Can you define something like `rememberAmberLauncher` which take `eventPublisher` as an argument or have a callback for success? The purpose would be to hide the implementation details.

app/src/main/kotlin/net/primal/android/core/utils/SignerUtils.kt
```diff
@@ -0,0 +1,16 @@
...
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+
+/* https://github.com/greenart7c3/Amber/blob/master/docs/Android.md */
+fun isExternalSignerInstalled(context: Context): Boolean {
+    val intent =
+        Intent().apply {
+            action = Intent.ACTION_VIEW
+            data = Uri.parse("nostrsigner:")
```
AleksandarIlic, line 12: Please use `String.toUri()` instead of `Uri.parse()`.

app/src/main/kotlin/net/primal/android/auth/login/LauncherExts.kt
```diff
@@ -0,0 +1,36 @@
+package net.primal.android.auth.login
+
+import android.content.Intent
+import androidx.activity.compose.ManagedActivityResultLauncher
+import androidx.activity.result.ActivityResult
+import androidx.core.net.toUri
+import net.primal.android.nostr.notary.NostrUnsignedEvent
+import net.primal.core.utils.serialization.CommonJson
+
+const val PACKAGE_NAME = "com.greenart7c3.nostrsigner"
```
AleksandarIlic, line 10: Should everything be public here?

AleksandarIlic: Looks like this file is related to Amber only. Maybe change the file name to reflect that. If it's not only for Amber, then maybe change the const name to say this is the package name of amber app.

app/src/main/kotlin/net/primal/android/auth/login/LoginContract.kt
```diff
@@ -10,20 +13,23 @@ interface LoginContract {
...
 
     sealed class UiEvent {
-        data object LoginRequestEvent : UiEvent()
+        data class LoginWithAmber(val pubkey: String) : UiEvent()
+        data class LoginRequestEvent(val nostrEvent: NostrEvent? = null) : UiEvent()
         data class UpdateLoginInput(val newInput: String) : UiEvent()
     }
 
     sealed class SideEffect {
+        data class RequestSign(val event: NostrUnsignedEvent) : SideEffect()
```
AleksandarIlic, line 32: I don't like the idea ViewModels asking UI to sign something. The UI should already be aware what needs to be signed and send it immeditally to avoid this back and forth.

app/src/main/kotlin/net/primal/android/auth/repository/LoginHandler.kt
```diff
@@ -21,38 +23,43 @@ class LoginHandler @Inject constructor(
     private val credentialsStore: CredentialsStore,
     private val nostrNotary: NostrNotary,
 ) {
-    enum class LoginType {
-        Npub,
-        Nsec,
-    }
-
-    suspend fun login(nostrKey: String, loginType: LoginType) =
+    suspend fun login(nostrKey: String, loginType: LoginType, authorizationEvent: NostrEvent?) =
```
AleksandarIlic, line 26: Does it make sense if we break this into `loginWithPrivateKey`, `loginWithPublicKey` and `loginWithRemoteSigner`?

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/auth/repository/LoginHandler.kt
markocic: I am not sure what is the best course of action here. If I haven't missed anything, it comes down to choice between having duplicated logic or a somewhat messy flow. I have no problem with implementing either. Let me know which one would you like!

app/src/main/kotlin/net/primal/android/signer/AmberLauncher.kt
```diff
@@ -0,0 +1,80 @@
...
+
+@Composable
+private fun rememberAmberLauncher(onResult: (ActivityResult) -> Unit): AmberLauncher =
+    rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.StartActivityForResult(),
+        onResult = onResult,
+    )
+
+/**
+ * Assumes `get_public_key` method has been called.  Extracts pubkey from it and
```
AleksandarIlic, line 29: `get_public_key` is related to the Amber. The docs should be in the context of our app. "Assumes `launchGetPublicKey`... " would be a better choice.

app/src/main/kotlin/net/primal/android/signer/event/EventBuilders.kt
```diff
@@ -0,0 +1,20 @@
...
+import net.primal.domain.nostr.asIdentifierTag
+
+
+fun buildAppSpecificDataEvent(pubkey: String) =
+    NostrUnsignedEvent(
+        pubKey = pubkey.extractProfileId() ?: "",
+        kind = NostrEventKind.ApplicationSpecificData.value,
+        tags = listOf("${UserAgentProvider.APP_NAME} App".asIdentifierTag()),
+        content = CommonJson.encodeToString(
+            AppSettingsDescription(description = "Sync app settings"),
```
AleksandarIlic, line 18: "Sync app settings" is probably going to be translated along with the app. In this case we should pass this text through an argument where this could be a default argument for now.

AleksandarIlic (approved):

AleksandarIlic (approved):

AleksandarIlic (approved):

PR #377, opened 2025-03-24
Fix crash NoSuchElementException in NoteEditorViewModel

mehmedalijaK:


AleksandarIlic (changes requested):
`referencedNoteId` in `NoteEditorArgs` should be `referencedNoteNevent`. With this change we are fixing the crash problem. Please update it and do the follow-up changes in the app.

app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
```diff
@@ -147,7 +147,9 @@ fun NoteEditorScreen(
                         contentPadding = PaddingValues(horizontal = 24.dp, vertical = 0.dp),
                         enabled = !state.publishing && !state.uploadingAttachments &&
                             state.attachments.none { it.uploadError != null } &&
-                            (state.content.text.isNotBlank() || state.attachments.isNotEmpty()),
+                            (state.content.text.isNotBlank() || state.attachments.isNotEmpty())
+                            && !state.fetchingNoteThread,
+                        loading = state.fetchingNoteThread,
```
markocic, line 152: I don't think we need loading state anymore. We now have `referencedNoteNevent` in the `args` already. This was done to prevent user from publishing invalid quote but now that is not a possibility.

app/src/main/kotlin/net/primal/android/editor/NoteEditorViewModel.kt
```diff
@@ -212,6 +212,7 @@ class NoteEditorViewModel @AssistedInject constructor(
                 .collect { conversation ->
                     val replyToNoteIndex = conversation.indexOfFirst { it.postId == replyToNoteId }
                     val thread = conversation.subList(0, replyToNoteIndex + 1)
+                    Timber.tag("replyingCheck").i(thread.toString())
```
markocic, line 215: We shouldn't have debug logs like this.

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -314,11 +314,13 @@ fun accountSwitcherCallbacksHandler(navController: NavController) =
...
-        onNoteReplyClick = { noteId -> navController.navigateToNoteEditor(NoteEditorArgs(referencedNoteId = noteId)) },
-        onNoteQuoteClick = { noteId ->
+        onNoteReplyClick = { referencedNoteId ->
+            navController.navigateToNoteEditor(NoteEditorArgs(referencedNoteNevent = referencedNoteId))
+        },
+        onNoteQuoteClick = { referencedNoteId ->
             navController.navigateToNoteEditor(
                 args = NoteEditorArgs(
-                    referencedNoteId = noteId,
+                    referencedNoteNevent = referencedNoteId,
```
markocic, line 323: This line is confusing. Rename this to `noteNevent`. Applies to the function above as well.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/events/NoteCallbacks.kt
```diff
@@ -2,8 +2,8 @@ package net.primal.android.notes.feed.note.ui.events
 
 data class NoteCallbacks(
     val onNoteClick: ((noteId: String) -> Unit)? = null,
-    val onNoteReplyClick: ((noteId: String) -> Unit)? = null,
-    val onNoteQuoteClick: ((noteId: String) -> Unit)? = null,
+    val onNoteReplyClick: ((referencedNoteId: String) -> Unit)? = null,
```
markocic, line 5: Having `NoteId` in the name is confusing. We are using `Nevent` here. Can you rename this to `noteNevent` to follow the naming convention we have already established? This goes for the line below as well.

app/src/main/kotlin/net/primal/android/notifications/list/ui/NotificationListItem.kt
```diff
@@ -102,7 +105,14 @@ fun NotificationListItem(
         onPostAction = { postAction ->
             if (postData != null) {
                 when (postAction) {
-                    FeedPostAction.Reply -> if (onReplyClick != null) onReplyClick(postData.postId)
+                    FeedPostAction.Reply -> if (onReplyClick != null) onReplyClick(
```
markocic, line 108: Could you use `onReplyClick?.invoke()` here to match the lines below? I think we have `PostData.toNevent()` extension function that does exactly this conversion to `Nevent`. Use that function where applicable.

app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
```diff
@@ -174,6 +176,8 @@ private fun NoteEditorContract.UiState.resolvePublishNoteButtonText() =
     when {
         publishing -> if (isReply) {
             stringResource(id = R.string.note_editor_reply_publishing_button)
+        } else if (isQuoting) {
```
markocic, line 179: This logic should check first for `isQuoting` since that is coming directly from "us". After that we should check if it's a reply. Switch these two and check if the `isReply` you modified can be simplified. If so, we should get rid of redundant `quotedNoteId` in state.

markocic (changes requested):
Good work. We should simplify some things and make the logic clearer.

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -314,11 +314,13 @@ fun accountSwitcherCallbacksHandler(navController: NavController) =
 fun noteCallbacksHandler(navController: NavController) =
     NoteCallbacks(
         onNoteClick = { noteId -> navController.navigateToThread(noteId = noteId) },
-        onNoteReplyClick = { noteId -> navController.navigateToNoteEditor(NoteEditorArgs(referencedNoteId = noteId)) },
-        onNoteQuoteClick = { noteId ->
+        onNoteReplyClick = { referencedNoteId ->
```
AleksandarIlic, line 317: `referencedNoteId` is probably a typo? I think we are receiving nevent here.

app/src/main/kotlin/net/primal/android/thread/notes/ThreadScreen.kt
```diff
@@ -238,7 +249,7 @@ fun ThreadScreen(
                     onExpandReply = { mediaUris ->
                         onExpandReply(
                             NoteEditorArgs(
-                                referencedNoteId = state.highlightPostId,
+                                referencedNoteNevent = state.highlightPostId,
```
AleksandarIlic, line 252: We should not put `state.highlightPostId` into nevent string. It's probably going to cause problems. Please fix to send proper nevent.

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/notes/feed/note/FeedNoteCard.kt
```diff
@@ -220,7 +223,17 @@ private fun FeedNoteCard(
                     ),
                 )
             },
-            onPostQuoteClick = { noteCallbacks.onNoteQuoteClick?.invoke(data.postId) },
+            onPostQuoteClick = {
+                noteCallbacks.onNoteQuoteClick?.invoke(
+                    Nevent(
```
markocic, line 228: I have noticed that we now create `Nevent` by hand in quite a number of places. This got me wondering if we can extract this and, as far as I can see, we can since all of the `Nevent`s are created based on `FeedPostUi`. Could you create an ext function `FeedPostUi.toNeventString()` and replace the usages where applicable?

markocic (changes requested):
I had a thought on how we can further improve this :smile:

markocic (dismissed):
Looks great. Thanks!

AleksandarIlic (approved):

app/src/main/kotlin/net/primal/android/notes/feed/model/FeedPostUi.kt
```diff
@@ -40,6 +45,7 @@ data class FeedPostUi(
     val eventZaps: List<EventZapUiModel> = emptyList(),
     val authorLegendaryCustomization: LegendaryCustomization? = null,
     val authorBlossoms: List<String> = emptyList(),
+    val eventRelayHints: EventRelayHints? = null,
```
AleksandarIlic, line 48: This should be `List<String> = emptyList()`

AleksandarIlic (approved):

PR #378, opened 2025-03-25
Feature/amber event signing

markocic:
Implements Amber signing events.  Uses `ContentResolver` to communicate with Amber. Because of this user is required to choose "I fully trust this application" when Amber prompts him on sign in.

### DIFF
--- a/app/src/main/kotlin/net/primal/android/nostr/publish/NostrPublisher.kt
+++ b/app/src/main/kotlin/net/primal/android/nostr/publish/NostrPublisher.kt
@@ -14,7 +14,6 @@ import net.primal.domain.nostr.NostrUnsignedEvent
 import net.primal.domain.publisher.NostrEventImporter
 import net.primal.domain.publisher.PrimalPublishResult
 import net.primal.domain.publisher.PrimalPublisher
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/net/primal/android/signer/SignerMethod.kt
```diff
@@ -1,6 +1,34 @@
...
+
+@Serializable(with = SignerMethod.Companion::class)
 enum class SignerMethod(val method: String) {
     GET_PUBLIC_KEY("get_public_key"),
     SIGN_EVENT("sign_event"),
+    NIP04_DECRYPT("nip04_decrypt"),
+    NIP04_ENCRYPT("nip04_encrypt"),
+    ;
+
+    companion object : KSerializer<SignerMethod> {
```
AleksandarIlic, line 20: This can go in it's own file. See `WalletKycLevelSerializer`.

AleksandarIlic (changes requested):
Code looks good, except two minor comments. I'm trying the build right now.

AleksandarIlic (approved):

PR #380, opened 2025-03-26
Fix broken thread line on Threads screen

mehmedalijaK:


AleksandarIlic (commented):
Interesting solution. We should probably reconsider where to draw the line to avoid the drawing on two places. Also, if this is a copy-paste code, then we should centralize the logic in place.

AleksandarIlic (approved):

PR #387, opened 2025-03-30
Handle `nostrUris` when parsing Uris

markocic:
Move ConversionUtils to `core:utils`;
Create Constants file;

Create NostrUriUtils in `core:utils` and move nostrUri logic there
    Not all logic could be moved. Some of it depends on Nip19. That was
    left in `NostrResources` and should be reconsidered.

### DIFF
--- a/core/utils/src/commonMain/kotlin/net/primal/core/utils/ConversionUtils.kt
+++ b/core/utils/src/commonMain/kotlin/net/primal/core/utils/ConversionUtils.kt
@@ -1,4 +1,4 @@
-package net.primal.domain.nostr.cryptography
+package net.primal.core.utils
 
 import fr.acinq.bitcoin.Bech32
 import fr.acinq.secp256k1.Hex
--- a/core/utils/src/commonMain/kotlin/net/primal/core/utils/NostrUriUtils.kt
+++ b/core/utils/src/commonMain/kotlin/net/primal/core/utils/NostrUriUtils.kt
@@ -49,7 +49,7 @@ fun String.isNostrUri(): Boolean {
         uri.startsWith(NEVENT) || uri.startsWith(NPROFILE)
 }
 
-fun String.cleanNostrUris(): String =
+fun String.clearAtSignFromNostrUris(): String =
     this
         .replace("@$NOSTR", NOSTR)
         .replace("@$NPUB", NPUB)
--- a/core/utils/src/commonMain/kotlin/net/primal/core/utils/NostrUriUtils.kt
+++ b/core/utils/src/commonMain/kotlin/net/primal/core/utils/NostrUriUtils.kt
@@ -1,0 +1,83 @@
+package net.primal.core.utils
+
+import io.github.aakira.napier.Napier
+
+
+private val nostrUriRegexPattern = Regex(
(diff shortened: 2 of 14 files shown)


core/utils/src/commonMain/kotlin/net/primal/core/utils/NostrUriUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+}
+
+
+fun String.isNostrUri(): Boolean {
+    val uri = lowercase()
+    return uri.startsWith(NOSTR) || uri.startsWith(NPUB) || uri.startsWith(NOTE) ||
+        uri.startsWith(NEVENT) || uri.startsWith(NPROFILE)
+}
+
+fun String.cleanNostrUris(): String =
```
AleksandarIlic, line 52: This function is a bit ambiguous. If it's used just in one place, maybe better to put it in that place. It does not clear the nostr uris, it just removes the `@` prefix from nostr uris. Maybe rename it as well as `clearAtSignFromNostrUris()`.

core/utils/src/commonMain/kotlin/net/primal/core/utils/ConversionUtils.kt
```diff
@@ -1,4 +1,4 @@
-package net.primal.domain.nostr.cryptography
+package net.primal.core.utils
```
AleksandarIlic, line 1: Can we actually delete the file in `domain:nostr` and move here the original file from `app` module? This way we can keep the git history I think.

The file should be also then renamed as `CryptographyUtils`.

Regarding the `DecoderException` and the tests. We can not use the spongycastle lib in commonMain. Can we find some replacement using the existing cryptography libs? If this is not straightforward, then let's do this comment in separate task.

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/mappers/remote/NostrResources.kt
```diff
@@ -284,8 +219,7 @@ fun List<String>.mapAsReferencedNostrUriDO(
     val refNote = postIdToPostDataMap[refNoteId]
     val refPostAuthor = profileIdToProfileDataMap[refNote?.authorId]
 
-    // Rewire once Nip19 is implemented for KMP
```
AleksandarIlic, line 287: Great catch!

AleksandarIlic (changes requested):
Looks great! Few minor thoughts.

core/utils/src/commonMain/kotlin/net/primal/core/utils/ConversionUtils.kt
markocic: This file was moved here from `domain:nostr` and the history is intact. I think the migration regarding `DecoderException` is not that straightforward. It would require some time to research possible solutions and assess the full extent of the migration seeing as it is one of the more used utils. I think it's best to leave this as a separate task.

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/mappers/remote/NostrResources.kt
markocic: Thanks!

core/utils/src/commonMain/kotlin/net/primal/core/utils/ConversionUtils.kt
AleksandarIlic: Lets do it later. Later we can move the original and delete the copy.

AleksandarIlic (approved):

PR #388, opened 2025-03-30
Reimplement observe recent users

mehmedalijaK:


app/src/main/kotlin/net/primal/android/user/repository/UserRepository.kt
```diff
@@ -366,24 +368,23 @@ class UserRepository @Inject constructor(
...
-//                    if (profile.metadata != null) {
-//                        UserProfileSearchItem(
-//                            metadata = profile.metadata.asProfileDataDO(),
-//                            followersCount = profile.stats?.followers,
-//                        )
-//                    } else {
-//                        null
-//                    }
-                    null
+                    withContext(dispatchers.io()) {
```
AleksandarIlic, line 376: Can we have just two calls instead? One to get all profile datas and one to get all profile stats? And then merge them into `UserProfileSearchItem`.

Something like this:
```kotlin
// Extract the profile IDs from the recent interactions
val profileIds = recentProfiles.map { it.profileId }

// Fetch all profiles and stats in bulk
val profiles = profileRepository.findProfilesByIds(profileIds).associateBy { it.id }
val statsMap = profileRepository.findProfilesStatsByIds(profileIds).associateBy { it.profileId }

// Map the fetched data to your result
profileIds.mapNotNull { profileId ->
    profiles[profileId]?.let { profile ->
        val stats = statsMap[profileId]
        UserProfileSearchItem(
            metadata = profile,
            followersCount = stats?.followers
        )
    }
}
```

This should give us better performance over multiple db queries.

app/src/main/kotlin/net/primal/android/user/repository/UserRepository.kt
```diff
@@ -51,6 +52,7 @@ class UserRepository @Inject constructor(
     private val fileUploader: PrimalFileUploader,
     private val usersApi: UsersApi,
     private val nostrPublisher: NostrPublisher,
+    private val profileRepository: ProfileRepository,
```
AleksandarIlic, line 55: FYI: UserRepository is a mess currently. We are going to clean it as soon as we are done with current blockers.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #390, opened 2025-03-30
Maintenance/sign exception migration

markocic:


app/src/main/kotlin/net/primal/android/messages/security/Nip04MessageCipher.kt
```diff
@@ -35,7 +35,7 @@ class Nip04MessageCipher @Inject constructor(
      * @throws MissingPrivateKey If no private key is found for the user.
```
AleksandarIlic, line 35: Fix docs to match new names. We have to maintain the docs as well, that's why I don't write it. :-D

AleksandarIlic (changes requested):
Resolve conflicts the fix the mentioned KDocs.

PR #392, opened 2025-03-31
 Fix sockets not working properly on init

mehmedalijaK:


AleksandarIlic:
This is not a proper way to fix the issue. This PR is trying to fix the consequence instead of the cause which is creation of the socket client was done in separate coroutine causing the request to go through before the client was created. I'll fix this in different branch.

AleksandarIlic:
Actually, this was a good direction for NostrSocketClient, but not for PrimalApiClient. The final fixes were done in 5fea1ba624a0822db17d53428c2fede287451bcd and c51157c7b09faaa761a903e9cd2f0283ab7c0837 commits.

PR #393, opened 2025-04-01
Maintenance/message encrypt exception

markocic:


domain/primal/src/commonMain/kotlin/net/primal/domain/repository/ChatRepository.kt
```diff
@@ -23,6 +24,7 @@ interface ChatRepository {
 
     suspend fun markAllMessagesAsRead(authorization: NostrEvent)
 
+    @Throws(MessageEncryptException::class)
```
AleksandarIlic, line 27: Whenever we declare `@Throws` on suspend functions in Kotlin `commonMain`, we need to put `kotlin.coroutines.cancellation.CancellationException` in the list as well.

This is necessary only for `suspend` function in order that proper Objective-C code is generated later for iOS platform. It's one of those boiler-plate things that we need to take care of it.

I'm gonna cover this in a task later down the road, but you should be aware of this.

AleksandarIlic (approved):

PR #398, opened 2025-04-02
Replace use of CommonJson with encodeToJsonString and decodeFromJsonStringOrNull

mehmedalijaK:


app/src/main/kotlin/net/primal/android/wallet/activation/WalletActivationViewModel.kt
```diff
@@ -84,7 +84,7 @@ class WalletActivationViewModel @Inject constructor(
 
     private fun loadAllCountries() =
         viewModelScope.launch {
-            val allCountries = CommonJson.decodeFromString<Regions>(WalletRegionJson).mapToListOfCountries()
+            val allCountries = WalletRegionJson.decodeFromJsonStringOrNull<Regions>()!!.mapToListOfCountries()
```
AleksandarIlic, line 87: We should never use `!!` in the code. This is serious, never use `!!`.

Instead make `allCountries` nullable and then fallback to `emptyList` in the state when you set it.

app/src/main/kotlin/net/primal/android/wallet/api/WalletApiImpl.kt
```diff
@@ -200,8 +200,8 @@ class WalletApiImpl @Inject constructor(
         val transactionsEvent = result.findPrimalEvent(kind = NostrEventKind.PrimalWalletTransactions)
             ?: throw WssException("Missing or invalid content in response.")
 
-        val txJsonArray = CommonJson.decodeFromString<JsonArray>(transactionsEvent.content)
-        val transactions = txJsonArray.mapNotNull {
+        val txJsonArray = transactionsEvent.content.decodeFromJsonStringOrNull<JsonArray>()
+        val transactions = txJsonArray!!.mapNotNull {
```
AleksandarIlic, line 204: Never use `!!`.

Adjust above line to be:
```kotlin
val txJsonArray = transactionsEvent.content.decodeFromJsonStringOrNull<JsonArray>()
     ?: throw WssException("Invalid content in 10_000_304 event.")
```

AleksandarIlic (changes requested):
We should never use `!!` in the code.

AleksandarIlic (approved):
Please resolve the conflicts again.

AleksandarIlic (approved):

PR #399, opened 2025-04-02
Maintenance/remove user account fetcher

markocic:


domain/primal/src/commonMain/kotlin/net/primal/domain/repository/ProfileRepository.kt
```diff
@@ -20,7 +19,7 @@ interface ProfileRepository {
     fun observeProfileData(profileId: String): Flow<ProfileData>
     fun observeProfileData(profileIds: List<String>): Flow<List<ProfileData>>
     fun observeProfileStats(profileId: String): Flow<ProfileStats?>
-    suspend fun fetchProfile(profileId: String)
+    suspend fun fetchProfile(profileId: String): ProfileData?
```
AleksandarIlic, line 22: I'm not sure on this, but since we don't have yet any standards on the repository interface function naming, we can go for now with this until we introduce a clear rule set for repository interfaces.

AleksandarIlic (approved):

PR #402, opened 2025-04-03
Implement `TransactionProfileData` to address the issue of missing profile data in transactions

markocic:


app/src/main/kotlin/net/primal/android/wallet/repository/WalletRepository.kt
```diff
@@ -38,12 +42,39 @@ class WalletRepository @Inject constructor(
...
+    private fun latestTransactions(userId: String) =
         createTransactionsPager(userId) {
             usersDatabase.walletTransactions().latestTransactionsPagedByUserId(userId = userId)
         }.flow
 
-    fun findTransactionById(txId: String) = usersDatabase.walletTransactions().findTransactionById(txId = txId)
+    /*
+     *  TODO: this should become a use case.  Could be migrated out with `findTransactionByIdOrNull` to
+     *   form some Interactor/Handler to deal with merging transactions with profiles.
+     */
```
AleksandarIlic, line 53: I agree with the comment. `Handler` might be appropriate, handling the merging of tx and profiles.

At the moment, TODOs are blockers for the `dev` merge to`main` and we should not put non-blockers as TODOs.

I suggest either removing the comment or fixing the comment.

AleksandarIlic (changes requested):
All looks great, except the TODO leftover.

app/src/main/kotlin/net/primal/android/wallet/repository/WalletRepository.kt
markocic: Resolving this TODO would require migrating `WalletRepository` out of `app` module. Considering the other blocker TODOs you mentioned, I think removing the comment is more aligned with our goals right now.

AleksandarIlic (approved):

PR #409, opened 2025-04-08
Maintenance/throws repositories

markocic:


app/src/main/kotlin/net/primal/android/networking/relays/RelaysSocketManager.kt
```diff
@@ -2,7 +2,7 @@ package net.primal.android.networking.relays
 
 import javax.inject.Inject
 import javax.inject.Singleton
-import kotlinx.coroutines.CancellationException
+import kotlin.coroutines.cancellation.CancellationException
```
AleksandarIlic, line 5: We should use this exception only on kmp common code. Not on the android code in `app.

And also I checked the usage, we should not catch this exception. Just revert this file and we are gonna fix this later.

domain/primal/src/commonMain/kotlin/net/primal/domain/repository/ArticleRepository.kt
```diff
@@ -1,35 +1,43 @@
...
     )
 
+    @Throws(NetworkException::class, CancellationException::class)
     suspend fun fetchArticleHighlights(
         userId: String,
         articleId: String,
         articleAuthorId: String,
     )
 
+    @Throws(CancellationException::class)
```
AleksandarIlic, line 28: No need to add only this exception. The compiler adds it by default, but if we add `@Throws` manually, then we have to declare it. Please apply this to all cases in this PR.

domain/primal/src/commonMain/kotlin/net/primal/domain/repository/EventInteractionRepository.kt
```diff
@@ -1,17 +1,22 @@
...
 
+import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.serialization.json.JsonArray
+import net.primal.domain.nostr.cryptography.SignatureException
+import net.primal.domain.nostr.publisher.NostrPublishException
 import net.primal.domain.publisher.PrimalPublishResult
 
 interface EventInteractionRepository {
 
+    @Throws(SignatureException::class, NostrPublishException::class, CancellationException::class)
```
AleksandarIlic, line 11: Can we list the exceptions like this when there are 3 or more?
```kotlin
@Throws(
        SignatureException::class,
        NostrPublishException::class,
        CancellationException::class,
    )
```
Or does the formatter changes to one-liner?

I know that for 3 or more arguments it breaks it into multi-line.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #415, opened 2025-04-11
Feature/amber signing poc

markocic:


app/src/main/kotlin/net/primal/android/settings/notifications/NotificationsSettingsViewModel.kt
```diff
@@ -121,13 +130,26 @@ class NotificationsSettingsViewModel @Inject constructor(
...
+                    is SignResult.Rejected -> {
+                        Timber.w(signResult.error)
+                        setState { copy(signatureError = signResult.error.asSignatureUiError()) }
+                    }
+
+                    is SignResult.Signed -> {
+                        withContext(dispatcherProvider.io()) {
+                            settingsRepository.fetchAndPersistAppSettings(signResult.event)
+                        }
+                    }
```
AleksandarIlic, line 152: I guess the catch clause `SignatureException`is no longer needed?

app/src/main/kotlin/net/primal/android/signer/AmberContentResolver.kt
```diff
@@ -110,13 +108,20 @@ fun ContentResolver.signEventWithAmber(
...
         } else {
-            return null
+            AmberSignResult.Undecided
         }
-    }
+    } ?: AmberSignResult.Undecided
+
+sealed class AmberSignResult {
+    data class Signed(val nostrEvent: NostrEvent) : AmberSignResult()
+    data object Undecided : AmberSignResult()
```
AleksandarIlic, line 125: `Ignored` might be a better name. The sign request was neither accept nor rejects, it was ignored. Right?

AleksandarIlic (commented):
If we can centralize amber signing like this, then this is a jack-pot!

app/src/main/kotlin/net/primal/android/signer/AmberContentResolver.kt
markocic: Getting "Undecided" (null) from Amber means the user didn't select remember option (neither approve nor reject), user has logged out of his account or we have a malformed sign request.

Since the most common meaning will be that the user didn't select remember choice option I decided to name it "Undecided", since the user hasn't previously decided on this request and we have to explicitly ask him.

Unfortunately Amber doesn't let us differentiate between previously mentioned cases so we should go with the most common one.

domain/nostr/src/commonMain/kotlin/net/primal/domain/nostr/cryptography/utils/SignatureUtils.kt
```diff
@@ -0,0 +1,35 @@
+package net.primal.domain.nostr.cryptography.utils
+
+import net.primal.domain.nostr.NostrEvent
+import net.primal.domain.nostr.cryptography.SignResult
+import net.primal.domain.nostr.cryptography.SignatureException
+
+fun SignResult.unwrapOrThrow(onFailure: ((SignatureException) -> Unit)? = null): NostrEvent =
```
AleksandarIlic, line 7: I had to look into the code to understand what is going on here. Maybe `unwrapSignedEventOrThrow`?

Or `NostrEvent` instead instead of `SignedEvent` might even be better.

This comment should apply to all functions in this file.

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
```diff
@@ -34,48 +45,83 @@ import net.primal.domain.nostr.cryptography.utils.toNpub
...
+            return SignResult.Rejected(error)
+        }
+
+        if (result != null) {
+            return SignResult.Signed(result)
+        }
+
+        signMutex.withLock {
+            setEffect(NotarySideEffect.RequestSignature(unsignedNostrEvent))
+            return _responses.receive()
```
AleksandarIlic, line 79: You could lift the `return` in front of the `if` and put the mutex in the `else` branch. And then you can remove ReturnCount warning. I think...

app/src/main/kotlin/net/primal/android/nostr/notary/NostrNotary.kt
```diff
@@ -34,48 +45,83 @@ import net.primal.domain.nostr.cryptography.utils.toNpub
...
         throw NotImplementedError()
     }
 
+    fun onSuccess(nostrEvent: NostrEvent) {
+        setResponse(SignResult.Signed(nostrEvent))
+    }
+
+    fun onFailure() {
+        setResponse(SignResult.Rejected(SigningRejectedException()))
+    }
```
AleksandarIlic, line 93: I'm not so sure about the names for this `onSuccess` and `onFailure` functions. But I don't have any suggestion at the moment.

AleksandarIlic (changes requested):
Beautiful solution!!!

PR #417, opened 2025-04-14
Implement remaining blossom api endpoints & Primal's blossom upload flow

mehmedalijaK:


app/src/main/kotlin/net/primal/android/networking/upload/PrimalUploadService.kt
```diff
@@ -101,21 +106,51 @@ class PrimalUploadService @Inject constructor(
             val base64Encoded = jsonPayload.encodeUtf8().base64()
             val authorizationHeader = "Nostr $base64Encoded"
 
-            val userBlossomServers = blossomRepository.getBlossomServers(userId).mapNotNull {
+            val blossomApis = blossomRepository.getBlossomServers(userId).mapNotNull {
                 runCatching { BlossomApiFactory.create(baseBlossomUrl = it) }.getOrNull()
             }.ifEmpty {
-                throw UnsuccessfulBlossomUpload(cause = IllegalStateException("Invalid blossom server list."))
+                listOf(blossomApi)
```
AleksandarIlic, line 112: We should actually throw an exception.  And `blossomApi` field should be deleted. If it's just for testing, then OK.

PR #418, opened 2025-04-14
Implement suggested blossom servers api

mehmedalijaK:


app/src/main/kotlin/net/primal/android/user/repository/BlossomRepository.kt
```diff
@@ -53,9 +57,15 @@ class BlossomRepository @Inject constructor(
...
-    suspend fun fetchSuggestedBlossomList(): List<String> {
-        return listOf("https://blossom.primal.net", "https://blossom.nostr.com")
-    }
+    suspend fun fetchSuggestedBlossomList(): List<String> =
+        withContext(dispatcherProvider.io()) {
+            try {
+                usersApi.getRecommendedBlossomServers()
+            } catch (error: WssException) {
+                Timber.w(error)
+                DEFAULT_BLOSSOM_LIST
```
AleksandarIlic, line 66: Default Blossom List and Suggested Blossom List are different lists. Just propagate the exception to the caller. You are already handling `WssException` in ViewModel.

If this endpoint fails, then we will have empty list of suggested servers and that's OK.

AleksandarIlic (approved):

PR #419, opened 2025-04-14
Implement publishing default blossom server list if existing user has no list

mehmedalijaK:


app/src/main/kotlin/net/primal/android/networking/upload/PrimalUploadService.kt
```diff
@@ -82,6 +82,8 @@ class PrimalUploadService @Inject constructor(
         onProgress: ((uploadedBytes: Int, totalBytes: Int) -> Unit)? = null,
     ): UploadResult =
         withContext(dispatchers.io()) {
+            blossomRepository.ensureBlossomServerList(userId = userId)
```
AleksandarIlic, line 85: Please update the `ensureBlossomServerList` to return the final blossom list. And then instead of using `blossomRepository.getBlossomServers(userId)` we can just use `ensure` functions.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #423, opened 2025-04-16
Stabilize feed ViewModel to prevent UI flicker

mehmedalijaK:


app/src/main/kotlin/net/primal/android/feeds/dvm/ui/DvmHeaderAndFeedList.kt
```diff
@@ -50,7 +50,6 @@ fun DvmHeaderAndFeedList(
                 NoteFeedList(
                     feedSpec = dvmFeed.data.buildSpec(specKind = feedSpecKind),
                     noteCallbacks = NoteCallbacks(),
-                    previewMode = true,
```
AleksandarIlic, line 53: This was here for a reason - to prevent users to open threads because we were unable to remember the scroll position in this mode. Did you fixed that?

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -85,7 +84,10 @@ fun NoteFeedList(
     header: @Composable (LazyItemScope.() -> Unit)? = null,
     stickyHeader: @Composable (LazyItemScope.() -> Unit)? = null,
 ) {
-    val viewModelKey by remember { mutableStateOf(if (!previewMode) feedSpec else UUID.randomUUID().toString()) }
```
AleksandarIlic, line 88: I forgot why this was added. It was a fix for something. Can you check if this was not broken?

AleksandarIlic (commented):
I don't understand what was the cause of the flickering and how does the fix works.

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -85,7 +85,14 @@ fun NoteFeedList(
...
+    val viewModelKey = remember(previewMode, feedSpec.toString()) {
+        if (!previewMode) {
+            "NoteFeedViewModel_$feedSpec"
```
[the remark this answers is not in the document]
mehmedalijaK: It's working with the prefix because it forces a new, unique String to be created every time. In contrast, when using feedSpec alone, Compose may cache the value with remember, even if feedSpec changes.

As a result, if we select the same feedSpec as before (even if it's technically a new instance), Compose might not re-run the block — so the viewModelKey stays the same, and the ViewModel doesn't fetch new data.

PR #424, opened 2025-04-16
Feature/extend muted list types

markocic:


AleksandarIlic (approved):
Everything looks great. We should do some feed performance testing in the future. We have no idea how fed performance is impacted by changing the SQL queries.

And I didn't like `MutedItemData.item` name, but I don't have any better alternative.

PR #428, opened 2025-04-17
Support multiple blossom servers

mehmedalijaK:


app/src/main/kotlin/net/primal/android/settings/media/MediaUploadsSettingsScreen.kt
```diff
@@ -83,6 +100,18 @@ private fun MediaUploadsSettingsScreen(
         }
     }
 
+    LaunchedEffect(state.error) {
```
AleksandarIlic, line 103: You should use our internal `SnackbarErrorHandler` for this.

app/src/main/kotlin/net/primal/android/settings/media/MediaUploadsSettingsScreen.kt
```diff
@@ -316,7 +348,7 @@ private fun LazyListScope.blossomMainServerInputItem(
             },
             showSupportContent = state.mode == MediaUploadsMode.View,
             buttonEnabled = state.newBlossomServerUrl != state.blossomServerUrl &&
-                state.newBlossomServerUrl != state.blossomServerMirrorUrl &&
+                !state.mirrorBlossomServerUrls.contains(state.newBlossomServerUrl) &&
```
AleksandarIlic, line 351: This `buttonEnabled` logic should be extracted into a `private fun` and reused in this file. I see there are similar implementation, only one argument is different.

AleksandarIlic (changes requested):
Looks great! Just two minor improvements necessary before we can merge.

PR #429, opened 2025-04-22
Fix adding word or hashtag to mute list fails, the value is lost

mehmedalijaK:


AleksandarIlic (approved):
Great work with fix and with the `MutedSettingsPager`.

PR #437, opened 2025-04-24
Implement topological sort for `FeedPost`

markocic:
We have relied on `createdAt` timestamp to figure out the root post and correct chain of replies. We have seen some cases in which root post doesn't have oldest timestamp which renders this method obsolete. This implements topological sort that follows `reply` and `root` tags on posts to figure out the correct order. This has the same time complexity as the DFS and shouldn't cause performance issues.

### DIFF
--- a/data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/feed/FeedRepositoryImpl.kt
+++ b/data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/feed/FeedRepositoryImpl.kt
@@ -25,6 +25,7 @@ import net.primal.data.repository.feed.processors.persistNoteRepliesAndArticleCo
 import net.primal.data.repository.feed.processors.persistToDatabaseAsTransaction
 import net.primal.data.repository.mappers.local.mapAsFeedPostDO
 import net.primal.data.repository.mappers.remote.asFeedPageSnapshot
+import net.primal.data.repository.utils.performTopologicalSort
 import net.primal.domain.common.exception.NetworkException
 import net.primal.domain.feeds.supportsNoteReposts
 import net.primal.domain.posts.FeedPageSnapshot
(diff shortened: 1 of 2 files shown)


data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/utils/ConversationUtils.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+        post.tags.find { it.hasRootMarker() }?.getTagValueOrNull()?.let {
+            adjacencyMap.getOrPut(key = it) { mutableSetOf() }
+                .add(post.eventId)
+        }
+    }
+
+    fun visit(node: FeedPost) {
+        if (permanentMark.contains(node.eventId)) return
+        if (temporaryMark.contains(node.eventId)) error("Impossible. Graph has cycle.")
```
AleksandarIlic, line 35: Noup. We are not going to crash the app because something went wrong here. If something goes wrong or invalid data, just return the original list.

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/feed/FeedRepositoryImpl.kt
```diff
@@ -152,6 +153,7 @@ internal class FeedRepositoryImpl(
             postId = noteId,
             userId = userId,
         ).map { list -> list.map { it.mapAsFeedPostDO() } }
+            .map { posts -> posts.performTopologicalSort() }
```
AleksandarIlic, line 156: Could we do
```kotlin
list.map { it.mapAsFeedPostDO() }
    .performTopologicalSort()
```
instead? To avoid two maps?

AleksandarIlic (changes requested):

PR #438, opened 2025-04-24
Fix muted users appearing in new posts pill

mehmedalijaK:


markocic (changes requested):

markocic (approved):
Looks good! Thanks.

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedViewModel.kt
```diff
@@ -79,10 +85,33 @@ class NoteFeedViewModel @AssistedInject constructor(
...
+        fetchLatestMuteList()
         subscribeToEvents()
         observeActiveAccount()
+        observeMutedUsers()
     }
 
+    private fun fetchLatestMuteList() =
+        viewModelScope.launch {
+            try {
+                withContext(dispatcherProvider.io()) {
```
AleksandarIlic, line 97: We should not have `dispatcherProvider` in viewModels. Instead make sure that `fetchAndPersistMuteList` is safe to call (it's using dispatcher provider in repository). I think it already does, but check it.

AleksandarIlic: Actually, let's not call fetch in this view model at all. Add it better to `UserDataUpdater`.

AleksandarIlic (changes requested):

PR #442, opened 2025-04-24
Feature/request delete note

markocic:


data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/events/EventInteractionRepositoryImpl.kt
```diff
@@ -103,6 +104,29 @@ class EventInteractionRepositoryImpl(
...
+                pubKey = userId,
+                kind = NostrEventKind.EventDeletion.value,
+                tags = listOf(
+                    noteId.asEventIdTag(authorPubkey = userId, relayHint = relayHint),
+                    eventKind.asKindTag(),
+                ),
+                content = content,
+            ),
+        )
+        database.posts().deletePostById(postId = noteId)
```
AleksandarIlic, line 125: This repo is handling events and here we are deleting post. What if in the future someone wants to delete some other event which is not the post?

AleksandarIlic (approved):

PR #444, opened 2025-04-25
Fix reporting and muting of own content

mehmedalijaK:


markocic (approved):
Works great!

Nitpick: not sure if we should put `activeAccountUserId` in states. I have done it first but now I am doubting it. It might be better to have `isActiveUserAuthor` or something like that in state.

AleksandarIlic (commented):
Is this working properly when you change the active account?

app/src/main/kotlin/net/primal/android/thread/notes/ThreadViewModel.kt
```diff
@@ -43,7 +43,10 @@ class ThreadViewModel @Inject constructor(
 
     private val highlightPostId = savedStateHandle.noteIdOrThrow.resolveNoteIdOrThrow()
 
-    private val _state = MutableStateFlow(UiState(highlightPostId = highlightPostId))
+    private val _state =
+        MutableStateFlow(
+            UiState(highlightPostId = highlightPostId, activeAccountUserId = activeAccountStore.activeUserId()),
```
AleksandarIlic, line 48: We should have an observer here.

app/src/main/kotlin/net/primal/android/thread/articles/details/ArticleDetailsViewModel.kt
```diff
@@ -73,7 +73,7 @@ class ArticleDetailsViewModel @Inject constructor(
     private val zapHandler: ZapHandler,
 ) : ViewModel() {
 
-    private val _state = MutableStateFlow(UiState())
+    private val _state = MutableStateFlow(UiState(activeAccountUserId = activeAccountStore.activeUserId()))
```
AleksandarIlic, line 76: We should have an observer here.

app/src/main/kotlin/net/primal/android/thread/articles/ArticleViewModel.kt
```diff
@@ -33,7 +33,7 @@ class ArticleViewModel @Inject constructor(
     private val bookmarksRepository: PublicBookmarksRepository,
 ) : ViewModel() {
 
-    private val _state = MutableStateFlow(UiState())
+    private val _state = MutableStateFlow(UiState(activeAccountUserId = activeAccountStore.activeUserId()))
```
AleksandarIlic, line 36: We should have an observer here.

AleksandarIlic (changes requested):
Update 4 viewmodels to observer active user account id instead of fetching it in the init.

AleksandarIlic (approved):

PR #445, opened 2025-04-25
Fix image gallery gap

mehmedalijaK:


AleksandarIlic (changes requested):
Can we have this value as `private const val GalleryGapSpace = 1.dp` in the file so we can easily change the spaces at one place in the future?

AleksandarIlic (approved):

PR #446, opened 2025-04-28
Implement push notifications option in Wallet settings

mehmedalijaK:


AleksandarIlic (changes requested):
This is really great implementation - good job! The only issue is that the this setting does not belong to notifications settings, but to the wallet settings. This is why the impl. looks weird. I have to check with backend team to see what can be done about it.

AleksandarIlic:
The requirements have changed. Closing this pull request since it's going to be few weeks until we get more details.

PR #452, opened 2025-04-30
Fix links not rendering on Android

mehmedalijaK:


app/src/main/kotlin/net/primal/android/wallet/transactions/details/TransactionDetailsScreen.kt
```diff
@@ -480,7 +480,7 @@ private fun TransactionNoteText(note: String) {
 
 @Composable
 private fun refineContentAsAnnotatedString(content: String): AnnotatedString {
-    val uriLinks = content.parseUris()
+    val uriLinks = content.detectUrls()
```
AleksandarIlic, line 483: `parseUris()` had default value to parse nostr uris as well. Is this necessary actually? If yes, we need to add that additionally. In the common code it was extracted into some other function, forgot the exact name.

AleksandarIlic (approved):

PR #453, opened 2025-04-30
Close `ThreadScreen` if the deleted note is root or ancestor note

markocic:


app/src/main/kotlin/net/primal/android/notes/feed/note/FeedNoteCard.kt
```diff
@@ -115,6 +116,14 @@ fun FeedNoteCard(
         viewModel.setEvent(UiEvent.DismissError)
     }
 
+    LaunchedEffect(viewModel, viewModel.effects) {
```
AleksandarIlic, line 119: 2nd param should be `onNoteDeleted`

AleksandarIlic (approved):

PR #456, opened 2025-04-30
Add support for BUD-08 in blossom uploader

mehmedalijaK:


core/networking-upload/src/commonMain/kotlin/net/primal/core/networking/blossom/Nip94Metadata.kt
```diff
@@ -0,0 +1,19 @@
+package net.primal.core.networking.blossom
```
AleksandarIlic, line 1: This belongs to `domain:nostr`

AleksandarIlic (changes requested):

PR #459, opened 2025-05-05
Implement loading and error states in ThreadScreen

mehmedalijaK:


app/src/main/kotlin/net/primal/android/thread/notes/ThreadViewModel.kt
```diff
@@ -154,6 +157,7 @@ class ThreadViewModel @Inject constructor(
                 }
             } catch (error: NetworkException) {
                 Timber.w(error)
+                setState { copy(error = UiError.FailedToFetchTopNoteZaps(error)) }
```
AleksandarIlic, line 160: We don't want to show an error if unable to fetch top zaps. Screen can work without this. Remove this from here and from `UiError` as well.

app/src/main/kotlin/net/primal/android/thread/notes/ThreadScreen.kt
```diff
@@ -305,14 +320,21 @@ private fun ThreadConversationLazyColumn(
...
+        } else {
+            ThreadLazyColumn(
+                modifier = Modifier.fillMaxSize(),
+                state = state,
+                onGoToWallet = onGoToWallet,
+                noteCallbacks = noteCallbacks,
+                onRootPostDeleted = onRootPostDeleted,
+                onUiError = onUiError,
+            )
+        }
```
AleksandarIlic, line 337: What if you try to open a note with deep-link, note which you don't have in the local database, and the fetching fails?

Compare this to the profile and article details for the same use case.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #461, opened 2025-05-07
Handle NetworkException during bookmark operations

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/feed/note/NoteViewModel.kt
```diff
@@ -347,6 +347,9 @@ class NoteViewModel @AssistedInject constructor(
             } catch (error: SigningRejectedException) {
                 setState { copy(error = UiError.NostrSignUnauthorized) }
                 Timber.w(error)
+            } catch (error: NetworkException) {
+                setState { copy(error = UiError.FailedToBookmarkNote(error)) }
+                Timber.w(error)
```
AleksandarIlic, line 352: Can you check all functions in this if they are catching `NetworkException`? And if they are not, check if they need to catch it or not and act accordingly.

mehmedalijaK: Yes, I’ve reviewed all the functions in this block. Functions that can throw NetworkException are already handled with a catch block.

AleksandarIlic (approved):

PR #464, opened 2025-05-08
Fix crash on onboarding next step beyond last enum value

mehmedalijaK:
Prevent IllegalArgumentException by safely handling out-of-bounds navigation in OnboardingStep.nextStep() and previousStep()


AleksandarIlic:
Were you able to reproduce this crash?

AleksandarIlic (approved):

PR #466, opened 2025-05-08
Fix crash by waiting for LazyList content before scrolling to top

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -159,6 +160,10 @@ private fun NoteFeedList(
 
     LaunchedEffect(shouldAnimateScrollToTop, state.shouldAnimateScrollToTop) {
         if (shouldAnimateScrollToTop || state.shouldAnimateScrollToTop == true) {
+            snapshotFlow { pagingItems.itemCount }
+                .filter { it > 0 }
+                .first()
```
AleksandarIlic, line 165: In `NoteFeedViewModel.showLatestNotes()`, there is an unsuccessful hack `delay(130.milliseconds)` to prevent this issue.

if you remove that hack delay, does your fix works?

AleksandarIlic (approved):

PR #472, opened 2025-05-13
Ensure selection only for first matching Zap option

mehmedalijaK:


AleksandarIlic (changes requested):
This does not work properly. Set 2 or more zap options to have the same amount, but different text. With this PR I can not select the 2nd zap option.

AleksandarIlic (approved):

PR #474, opened 2025-05-14
Fix zap insert constraint (EventZapDao.insert) crash

### LINKED ISSUE #141: Two fast zap clicks on a specific note make Primal crash
Two fast zap clicks on a specific note make Primal crash.
Note: note1dqecfx3dlqdw84se9lpfqfmqnzfr9csfgnjxanz5d4gxtwaz3h6quslpr9

https://github.com/PrimalHQ/primal-android-app/assets/1178856/953ca254-11bf-4861-a527-4fd731866a1f

mehmedalijaK:
Closes #141

### DIFF
--- a/data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/events/EventZap.kt
+++ b/data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/events/EventZap.kt
@@ -1,16 +1,11 @@
 package net.primal.data.local.dao.events
 
 import androidx.room.Entity
-import androidx.room.Index
 import androidx.room.PrimaryKey
 import net.primal.domain.links.CdnImage
 import net.primal.domain.premium.PrimalLegendProfile
 
-@Entity(
-    indices = [
-        Index(value = ["zapSenderId", "eventId", "zapRequestAt"]),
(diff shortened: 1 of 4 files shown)


AleksandarIlic (changes requested):
This would merge zaps with same `zapCreatedAt`. Let's better create an internal auto increment primary key instead of trying to do composite primary key from existing fields..

data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/events/EventZap.kt
```diff
@@ -1,17 +1,18 @@
...
 import net.primal.domain.links.CdnImage
 import net.primal.domain.premium.PrimalLegendProfile
 
 @Entity(
-    primaryKeys = [
-        "zapSenderId",
-        "eventId",
-        "zapRequestAt",
+    indices = [
+        Index(value = ["zapSenderId", "eventId", "zapRequestAt"]),
```
AleksandarIlic, line 11: I'm not sure if we should have this fields indexed? Any particular reason? If it's just because this was a Primary Key in the past, I think we can drop it.

PR #477, opened 2025-05-18
Ensure cap attachment preview width at 92 % (stops overflow, adds peek)

mehmedalijaK:


app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
```diff
@@ -631,3 +632,5 @@ private val avatarSizeDp = 42.dp
 private val connectionLineOffsetXDp = 40.dp
 private val attachmentsHeightDp = 160.dp
 private val avatarsColumnWidthDp = avatarSizeDp + 24.dp
+
+private const val ATTACHMENT_FILL_FRACTION = 0.92f
```
AleksandarIlic, line 636: This might not be true for some other devices screens. Is there a way to handle this differently? For example see how many `dp`s is max width and then use that value instead of percentage? `BoxWithConstraints` gives you option to get `maxWidth` and `maxHeight`.

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
```diff
@@ -527,7 +527,8 @@ private fun NoteAttachmentsLazyRow(
     LazyRow(
         modifier = modifier,
         contentPadding = PaddingValues(
-            start = avatarsColumnWidthDp - 8.dp,
+            start = 68.dp,
```
AleksandarIlic, line 530: We are going to have problem here if values at the bottom of this file change. Can you fix the `start` padding to be dynamically resolved?

PR #478, opened 2025-05-19
Preselect the correct tab in note reactions

mehmedalijaK:


app/src/main/kotlin/net/primal/android/navigation/NavigationArguments.kt
```diff
@@ -108,3 +109,9 @@ inline val SavedStateHandle.lnbc: String? get() = get<String>(LNBC)
 const val TRANSACTION_ID = "transactionId"
 inline val SavedStateHandle.transactionIdOrThrow: String
     get() = get(TRANSACTION_ID) ?: throw IllegalArgumentException("Missing required transactionId argument.")
+
+const val INITIAL_TAB = "initialTab"
```
AleksandarIlic, line 113: This is generic name, but the return type is specific to Reactions. Could you please rename it to be `initialReactionsTab` or similar? In this file and everywhere else where applicable in this PR.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/events/NoteCallbacks.kt
```diff
@@ -13,7 +13,8 @@ data class NoteCallbacks(
     val onHashtagClick: ((hashtag: String) -> Unit)? = null,
     val onMediaClick: ((event: MediaClickEvent) -> Unit)? = null,
     val onPayInvoiceClick: ((event: InvoicePayClickEvent) -> Unit)? = null,
-    val onEventReactionsClick: ((eventId: String) -> Unit)? = null,
+//    val onEventReactionsClick: ((eventId: String) -> Unit)? = null,
```
AleksandarIlic, line 16: This should be removed?

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/events/ReactionTab.kt
```diff
@@ -0,0 +1,3 @@
+package net.primal.android.notes.feed.note.ui.events
+
+enum class ReactionTab { ZAPS, LIKES, REPOSTS, REPLIES }
```
AleksandarIlic, line 3: This should not belong here and the name is coupled with UI.

You should put it in `domain:primal` module, call it `ReactionType` or something in that direction and use singular form for enum names (`ZAP`, `LIKE`, etc).

Then in the UI you are going to map this to the actual tab position. Then when some other parts of the app use this, they can do whatever they want with it (even if they don't have tabs).

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteStatsRow.kt
```diff
@@ -12,15 +13,21 @@ import androidx.compose.ui.text.font.FontWeight
...
 import net.primal.android.R
 import net.primal.android.notes.feed.model.EventStatsUi
+import net.primal.android.notes.feed.note.ui.events.ReactionTab
 import net.primal.android.theme.AppTheme
 
 @Composable
-fun ThreadNoteStatsRow(eventStats: EventStatsUi, modifier: Modifier = Modifier) {
+fun ThreadNoteStatsRow(
+    eventStats: EventStatsUi,
+    onStatClick: (ReactionTab) -> Unit,
```
AleksandarIlic, line 22: Maybe `onReactionTypeClick`?

app/src/main/kotlin/net/primal/android/navigation/NavigationArguments.kt
AleksandarIlic: Actually, I commented below that we should decoupled it from tabs. So that `ReactionType` comment applies for here as well.

AleksandarIlic (changes requested):
Good work. We just need to improve the architecture a bit.

AleksandarIlic (approved):

PR #480, opened 2025-05-19
Load the new notifications and scroll to top

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notifications/list/NotificationsScreen.kt
```diff
@@ -148,6 +138,21 @@ fun NotificationsScreen(
...
 
+    var previousUnseenNotificationIds by remember {
+        mutableStateOf<List<String>>(emptyList())
+    }
+
+    LaunchedEffect(canScrollUp, state.unseenNotifications) {
+        val currentIds = state.unseenNotifications.flatten().map { it.notificationId }
+
+        if (canScrollUp && currentIds != previousUnseenNotificationIds) {
+            previousUnseenNotificationIds = currentIds
```
AleksandarIlic, line 149: Why do you need to track seen notifications like this?

AleksandarIlic (approved):

PR #486, opened 2025-05-21
Implement `ConnectionIndicatorOverlay`

markocic:


app/src/main/kotlin/net/primal/android/core/compose/connectionindicator/ConnectionIndicatorOverlay.kt
```diff
@@ -0,0 +1,245 @@
...
+import net.primal.android.R
+import net.primal.android.core.compose.icons.PrimalIcons
+import net.primal.android.core.compose.icons.primaliconpack.NoConnectionBlack
+import net.primal.android.core.compose.icons.primaliconpack.NoConnectionWhite
+import net.primal.android.theme.AppTheme
+
+internal const val SHOW_NOTICE_DURATION_TIME_IN_MILLIS = 3000L
+
+@Composable
+fun ConnectionIndicatorOverlay(viewModel: ConnectionIndicatorViewModel, content: @Composable () -> Unit) {
```
AleksandarIlic, line 60: Maybe you can create viewModel here instead of passing it from `MainActivity`?

app/src/main/kotlin/net/primal/android/core/compose/connectionindicator/ConnectionIndicatorViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+    val state = _state.asStateFlow()
+    private fun setState(reducer: UiState.() -> UiState) = _state.getAndUpdate { it.reducer() }
+
+    init {
+        observeCachingServiceConnection()
+    }
+
+    private fun observeCachingServiceConnection() =
+        viewModelScope.launch {
+            primalApiClient.connectionStatus.collect {
```
AleksandarIlic, line 29: What happens when you are switching caching services? Does the notice shows up during reconnection? If yes, we should probably avoid. Maybe some back-off approach here. Forgot the exact name of the operator for that.

AleksandarIlic (commented):
Amazing!

AleksandarIlic (approved):

PR #489, opened 2025-05-22
Improve animated GIFs support

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/compose/UniversalAvatarThumbnail.kt
```diff
@@ -49,6 +49,7 @@ fun UniversalAvatarThumbnail(
     backgroundColor: Color = AppTheme.extraColorScheme.surfaceVariantAlt1,
     onClick: (() -> Unit)? = null,
     hasInnerBorderOverride: Boolean = true,
+    isAvatarAnimated: Boolean = false,
```
AleksandarIlic, line 52: Maybe better name would be `supportAnimation`? Here and in functions below.

AleksandarIlic (changes requested):
Great job fixing this. I have only small nitpick regarding variable name.

AleksandarIlic (approved):

PR #507, opened 2025-06-05
Feature/explore follow packs

markocic:


data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/explore/FollowPackProfileCrossRef.kt
```diff
@@ -0,0 +1,9 @@
+package net.primal.data.local.dao.explore
+
+import androidx.room.Entity
+
+@Entity(primaryKeys = ["followPackIdentifier", "userId"])
+data class FollowPackProfileCrossRef(
+    val followPackIdentifier: String,
```
AleksandarIlic: Are you sure we can only with `followPackIdentifier`? Identifiers are unique per user only. Maybe we need to include the id of the follow pack author to avoid scenario where two different users have the same identifiers.

data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/explore/FollowPackProfileCrossRef.kt
```diff
@@ -0,0 +1,9 @@
...
+    val userId: String,
```
AleksandarIlic: If I got this correctly, this should be `profileId`? It points which profile is on the above follow pack?

data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/explore/FollowPackData.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import androidx.room.Entity
+
+@Entity(primaryKeys = ["identifier", "createdBy"])
+data class FollowPackData(
+    val identifier: String,
+    val title: String,
+    val coverImage: String?,
+    val description: String?,
+    val createdBy: String,
```
AleksandarIlic: The name is ambiguous. Is it the name? Pubkey? Id? 
I assume we are storing the `id`, so it could be named `authorId`? We already have `authorId` as columns in some other tables.

data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/explore/FollowPackData.kt
```diff
@@ -0,0 +1,14 @@
+package net.primal.data.local.dao.explore
+
+import androidx.room.Entity
+
+@Entity(primaryKeys = ["identifier", "createdBy"])
+data class FollowPackData(
+    val identifier: String,
+    val title: String,
+    val coverImage: String?,
```
AleksandarIlic: Maybe `coverImageUrl`? I assume it's url?
Wait, why are we storing the original url? Maybe we should store `CdnImage` like we do for some other entities?

data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/explore/FollowPackRemoteKey.kt
```diff
@@ -0,0 +1,14 @@
+package net.primal.data.local.dao.explore
+
+import androidx.room.Entity
+
+@Entity(
+    primaryKeys = ["pubkey", "identifier"],
+)
+data class FollowPackRemoteKey(
+    val pubkey: String,
```
AleksandarIlic: I don't understand what is this `pubkey` for. It's like we just say `id` somewhere. Is this the `id` of the this remote key? If not, then we should name it to explain to whom id belongs to - `followPackId`? `authorId`? Or something else?

AleksandarIlic (commented):
Here are the comments for `local-caching` module.

data/remote-caching/src/commonMain/kotlin/net/primal/data/remote/api/explore/model/FollowPackRequestBody.kt
```diff
@@ -0,0 +1,9 @@
+package net.primal.data.remote.api.explore.model
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class FollowPackRequestBody(
+    val pubkey: String,
```
AleksandarIlic: Can we be precise here to what this pubkey points to? I assume `authorId`? If you remember we had the problem with caching api where it was not clear which pubkey do we send - this is the same situation for somebody else who is reading the code.

data/remote-caching/src/commonMain/kotlin/net/primal/data/remote/api/explore/ExploreApiImpl.kt
```diff
@@ -43,6 +47,52 @@ internal class ExploreApiImpl(
         )
     }
 
+    override suspend fun getFollowLists(body: FollowListsRequestBody): FollowListsResponse {
+        val queryResult = primalApiClient.query(
+            message = PrimalCacheFilter(
+                primalVerb = net.primal.data.remote.PrimalVerb.FOLLOW_LISTS.id,
+                optionsJson = CommonJsonImplicitNulls.encodeToString(body),
```
AleksandarIlic, line 54: We should use `body.encodeToJsonString()` if possible here and below?

Maybe you should adapt the body data class to avoid need for special implicit nulls serializer?

markocic: How could I achieve that? I need parameter "until" to be either `Long` or `null`. In case of `null` I don't want it sent at all. I could make another data class and use it instead but that just complicates logic in other places.

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/mappers/local/FollowPackMapper.kt
```diff
@@ -0,0 +1,53 @@
...
+    followersCountMap: Map<String, Int>,
+): FollowPackDO? {
+    return FollowPackDO(
+        identifier = identifier,
+        title = title,
+        coverImage = coverImage,
+        description = description,
+        createdByProfileId = createdBy,
+        createdBy = profiles[createdBy]
+            ?.asFollowPackProfileData(followersCount = followersCountMap[createdBy] ?: 0),
```
AleksandarIlic: nitpick: Better line break would be with on the arguments.

data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/explore/FollowPackProfileCrossRef.kt
markocic: I was aware of this. Problem is ae can't use `Relation` join as Jetpack Room doesn't allow for join on two columns. We could introduce a surrogate primary key that is auto generated for join purposes.

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/mappers/local/FollowPackMapper.kt
```diff
@@ -0,0 +1,53 @@
+package net.primal.data.repository.mappers.local
```
AleksandarIlic, line 1: Can all functions in this file be `internal`?

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/explore/FollowPackResponseProcessor.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    val followPacks = followListEvents
+        .orderByPagingIfNotNull(pagingEvent = pagingEvent)
+        .mapAsFollowPackData()
+
+    database.withTransaction {
+        database.profiles().insertOrUpdateAll(data = profiles)
+        database.followPacks().upsertFollowPackData(data = followPacks.map { it.first })
+        database.followPacks()
+            .clearCrossRefsByPackIdentifiers(identifiers = followPacks.map { it.first }.map { it.identifier })
```
AleksandarIlic: nitpick: Line break on arguments is easier to read.

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/explore/ExploreRepositoryImpl.kt
```diff
@@ -216,4 +269,25 @@ class ExploreRepositoryImpl(
         queryRemoteUsers {
             exploreApi.getPopularUsers()
         }
+
+    @OptIn(ExperimentalPagingApi::class)
+    private fun createPager(pagingSourceFactory: () -> PagingSource<Int, FollowPack>) =
```
AleksandarIlic: Could we rename it to `createFollowPacksPager`?  Or make `remoteMediator` as an argument and keep the name (then we can reuse this function when needed) in this repo later.

data/repository-caching/src/commonMain/kotlin/net/primal/data/repository/explore/paging/FollowPackMediator.kt
```diff
@@ -0,0 +1,177 @@
...
+import net.primal.domain.common.exception.NetworkException
+
+@ExperimentalPagingApi
+internal class FollowPackMediator(
+    private val exploreApi: ExploreApi,
+    private val database: PrimalDatabase,
+    private val dispatcherProvider: DispatcherProvider,
+) : RemoteMediator<Int, FollowPackPO>() {
+
+    private val lastRequests: MutableMap<LoadType, Pair<FollowListsRequestBody, Long>> = mutableMapOf()
```
AleksandarIlic: Idea: This keeps showing up in our mediators. Maybe we could create a typed class to handle this case and reuse it everywhere? It's just and idea - no need to work on this right now.

AleksandarIlic (commented):
Mostly nitpicks comments for `repository-caching`.

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/ExploreRepository.kt
```diff
@@ -1,18 +1,41 @@
...
 
+import androidx.paging.PagingData
+import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.coroutines.flow.Flow
 import net.primal.domain.common.UserProfileSearchItem
+import net.primal.domain.common.exception.NetworkException
 
 interface ExploreRepository {
 
+    @Throws(NetworkException::class, CancellationException::class)
```
AleksandarIlic, line 11: Thanks for fixing this!

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/FollowPack.kt
```diff
@@ -0,0 +1,13 @@
+package net.primal.domain.explore
+
+data class FollowPack(
+    val identifier: String,
+    val title: String,
+    val coverImage: String?,
+    val description: String?,
+    val createdByProfileId: String,
+    val createdBy: FollowPackProfileData?,
+    val peopleCount: Int,
```
AleksandarIlic: Is `peopleCount` really necessary? Can we do `people.size`?

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/FollowPack.kt
```diff
@@ -0,0 +1,13 @@
+package net.primal.domain.explore
+
+data class FollowPack(
+    val identifier: String,
+    val title: String,
+    val coverImage: String?,
+    val description: String?,
+    val createdByProfileId: String,
```
AleksandarIlic: What do you think of `authorId` and `authorProfileData`?

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/FollowPackProfileData.kt
```diff
@@ -0,0 +1,13 @@
+package net.primal.domain.explore
+
+import net.primal.domain.links.CdnImage
+import net.primal.domain.premium.PrimalPremiumInfo
+
+data class FollowPackProfileData(
```
AleksandarIlic, line 6: Can we reuse some other existing `ProfileData` (or extend it) instead of creating an additional entity? Does it make sense?

AleksandarIlic (commented):
Domain comments.

app/src/main/kotlin/net/primal/android/core/compose/LazyListStateUtils.kt
```diff
@@ -0,0 +1,35 @@
...
+ * is currently visible in the viewport of a lazy list (e.g. LazyColumn or LazyRow).
+ *
+ * This function creates a state that observes the list's visible items and tracks whether the item
+ * with the specified [key] is visible. It updates the visibility reactively whenever the visible items change.
+ *
+ * @param key The key of the item whose visibility should be tracked.
+ * @return `true` if the item with the given [key] is currently visible, `false` otherwise.
+ */
+@Composable
+fun LazyListState.isItemVisible(key: Any?): Boolean {
```
AleksandarIlic: This might be confusing. We should return `MutableState<Boolean>` and let caller use `by` or `.value`

Also it should be named `rememberIsItemVisible` or something like this.

app/src/main/kotlin/net/primal/android/core/compose/LazyListStateUtils.kt
```diff
@@ -0,0 +1,35 @@
...
+    val isVisible = remember {
+        mutableStateOf(layoutInfo.visibleItemsInfo.firstOrNull { it.key == key } != null)
+    }
+    LaunchedEffect(this, key) {
```
AleksandarIlic: Did you consider using `derivedStateOf`?

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/FollowPackProfileData.kt
markocic: This data class represents ProfileData with followers count from ProfileStats. Extending it would mess with ProfileData in other places as we would have joins everywhere or invalid data with followers being `null` or `0`.

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/FollowPack.kt
markocic: `people.size` represents array of profiles we have information on. Backend doesn't return full list of profiles thus making array size a bad indicator of people count in a follow pack. Real indicator is number of pubkey tags in follow pack nostr event which is what this number represents.

app/src/main/kotlin/net/primal/android/explore/home/followpack/FollowPackScreen.kt
```diff
@@ -0,0 +1,483 @@
...
+            ListNoContent(
+                modifier = Modifier.fillMaxSize(),
+                noContentText = stringResource(id = R.string.follow_pack_unable_to_load),
+                onRefresh = { eventPublisher(UiEvent.RefreshFollowPack) },
+            )
+        }
+    }
+}
+
+fun LazyListScope.followPackInfo(followPack: FollowPackUi, onProfileClick: (String) -> Unit) {
```
AleksandarIlic: Can this be `private`?  If not maybe put in `ui` package.

app/src/main/kotlin/net/primal/android/explore/home/people/model/FollowPackUi.kt
```diff
@@ -0,0 +1,36 @@
...
+    val description: String?,
+    val createdByProfileId: String,
+    val createdBy: UserProfileItemUi?,
+    val highlightedPeople: List<UserProfileItemUi>,
+    val peopleCount: Int,
+    val people: List<UserProfileItemUi>,
+    val updatedAt: Instant,
+)
+
+internal const val HIGHLIGHTED_PEOPLE_COUNT = 5
```
AleksandarIlic: This should be `private` I think so it's not visible across the entire `app` module.

app/src/main/kotlin/net/primal/android/explore/home/people/ExplorePeopleViewModel.kt
```diff
@@ -2,179 +2,26 @@ package net.primal.android.explore.home.people
...
-import timber.log.Timber
 
 @HiltViewModel
 class ExplorePeopleViewModel @Inject constructor(
     private val exploreRepository: ExploreRepository,
-    private val activeAccountStore: ActiveAccountStore,
-    private val profileRepository: ProfileRepository,
-    private val userRepository: UserRepository,
 ) : ViewModel() {
+    private fun buildFollowPacksPager() =
```
AleksandarIlic, line 23: Can we put functions after the MVI boilerplate and init block?

data/local-caching/src/commonMain/kotlin/net/primal/data/local/dao/explore/FollowPackProfileCrossRef.kt
AleksandarIlic: Let's go with `a` tag as unique id. If it's missing, we can recreate it.

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/FollowPack.kt
AleksandarIlic: Ok. Now makes sense. Let's go with `profile` term instead of `people`, there can be bots on the list, not just people.

domain/primal/src/commonMain/kotlin/net/primal/domain/explore/FollowPackProfileData.kt
AleksandarIlic: Ok. Let's keep it as it is.

data/remote-caching/src/commonMain/kotlin/net/primal/data/remote/api/explore/ExploreApiImpl.kt
AleksandarIlic: Ok. Let's keep it as it is.

AleksandarIlic (approved):

PR #512, opened 2025-06-11
Filter out Primal note URLs already represented as NostrUris

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteContent.kt
```diff
@@ -259,12 +259,19 @@ fun NoteContent(
         }
 
         if (data.uris.isNotEmpty()) {
+            val existingNostrUris = data.nostrUris
+                .filter { it.type == EventUriNostrType.Note }
+                .map { it.uri }
+                .toSet()
```
AleksandarIlic: Please use `remember` for calculations like this in Compose.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteContent.kt
```diff
@@ -259,12 +259,19 @@ fun NoteContent(
...
+
             NoteAttachments(
                 modifier = Modifier
                     .fillMaxWidth()
                     .padding(top = if (contentText.isEmpty()) 4.dp else 6.dp)
                     .heightIn(min = 0.dp, max = 500.dp),
-                eventUris = data.uris,
+                eventUris = data.uris.filterNot { uriItem ->
+                    uriItem.url.substringAfterLast("/") in existingNostrUris
```
AleksandarIlic: We want to remove only `primal.net` links if we already have a referenced event. Other links should show up for now.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteContent.kt
AleksandarIlic: Also, we should support both Notes and Articles.

AleksandarIlic (approved):

PR #528, opened 2025-06-28
Feature/macrobenchmark

markocic:


app/src/main/AndroidManifest.xml
```diff
@@ -45,17 +44,19 @@
         android:usesCleartextTraffic="true"
         tools:replace="android:allowBackup"
         tools:targetApi="33">
+        <profileable
```
AleksandarIlic: I'm not sure if this is safe for public production builds. Can we use specific `AndroidManifest.xml` just for benchmark flavour?

settings.gradle.kts
```diff
@@ -1,5 +1,7 @@
 import java.net.URI
 
+include(":macrobenchmark")
```
AleksandarIlic: Could you please move this to the bottom of the file where all other modules are? Or move others here. All modules should be at one place.

macrobenchmark/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+            value = "\"${configProperties?.getProperty("localStorage.benchmarkNsec", "")}\"",
+        )
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        testInstrumentationRunnerArguments["androidx.benchmark.suppressErrors"] = "EMULATOR"
+        testInstrumentationRunnerArguments["androidx.benchmark.killExistingPerfettoRecordings"] = "false"
+    }
+
+    buildTypes {
+        create("benchmark") {
```
AleksandarIlic, line 41: Why do we need this build type here and in `app` module?

macrobenchmark/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+            signingConfig = getByName("debug").signingConfig
+            matchingFallbacks += listOf("release")
+        }
+    }
+
+    buildFeatures {
+        buildConfig = true
+    }
+
+    flavorDimensions += listOf("distribution")
```
AleksandarIlic, line 52: Are flavours really necessary for this module? I think we are going to test performance only with `aosp` builds.

macrobenchmark/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    defaultConfig {
+        minSdk = 24
+        targetSdk = 36
+
+        buildConfigField(
+            type = "String",
+            name = "BENCHMARK_NSEC",
+            value = "\"${configProperties?.getProperty("localStorage.benchmarkNsec", "")}\"",
```
AleksandarIlic: Let's use just `benchmark.nsec`. The `localStorage` part is for local encryption.

macrobenchmark/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+    val configFile = File("config.properties")
+    if (configFile.exists()) {
+        Properties().apply { load(configFile.reader()) }
+    } else {
+        null
+    }
+}
+
+android {
+    namespace = "net.primal.macrobenchmark"
```
AleksandarIlic: nitpick: I would keep `net.primal.android` prefix and then add `macrobenchmark` if it needs to be different.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #529, opened 2025-06-30
Set `BigDecimal`s `DecimalMode` to US currency when dividing

markocic:
Default DecimalMode has unlimited `decimalPrecision` which causes a non-terminating result of division. `US_CURRENCY` mode uses biggest possible `decimalPrecision` which avoids this problem.

### DIFF
--- a/core/utils/src/commonMain/kotlin/net/primal/core/utils/CurrencyConversionUtils.kt
+++ b/core/utils/src/commonMain/kotlin/net/primal/core/utils/CurrencyConversionUtils.kt
@@ -1,6 +1,7 @@
 package net.primal.core.utils
 
 import com.ionspin.kotlin.bignum.decimal.BigDecimal
+import com.ionspin.kotlin.bignum.decimal.DecimalMode
 import com.ionspin.kotlin.bignum.decimal.toBigDecimal
 
 @Suppress("TooManyFunctions")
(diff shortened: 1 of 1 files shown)


core/utils/src/commonMain/kotlin/net/primal/core/utils/CurrencyConversionUtils.kt
```diff
@@ -40,7 +41,7 @@ object CurrencyConversionUtils {
 
     fun BigDecimal.fromUsdToSats(exchangeBtcUsdRate: Double?): ULong {
         val rate = exchangeBtcUsdRate ?: 0.0
-        val btcAmount = divide(rate.toBigDecimal())
+        val btcAmount = divide(rate.toBigDecimal(decimalMode = DecimalMode.US_CURRENCY))
```
AleksandarIlic, line 44: Should we use the same `decimalMode` for all `toBigDecimal` calls in this and/or other files files? Do we need maybe our own `toBigDecimal()`?

AleksandarIlic (changes requested):
Please check my comment above.

core/utils/src/commonMain/kotlin/net/primal/core/utils/CurrencyConversionUtils.kt
markocic: No, because problem described is only present when dividing. We don't have any other BigDecimal divisions.

AleksandarIlic (approved):

PR #530, opened 2025-06-30
Move `FocusRequester` closer to the `TextField` that uses it

markocic:
`FocusRequester` is initialized after the component that uses has been added to the UI tree. Moving it closer to the `TextField` and adding `awaitFrame()` before calling `requestFocus()` ensures that at the time of requesting focus we have initialized `FocusRequester`.

### DIFF
--- a/app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
+++ b/app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
@@ -66,6 +66,7 @@ import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import java.util.*
+import kotlinx.coroutines.android.awaitFrame
 import net.primal.android.R
 import net.primal.android.articles.feed.ui.FeedArticleListItem
 import net.primal.android.articles.feed.ui.FeedArticleUi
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
```diff
@@ -492,12 +487,17 @@ private fun NoteEditor(
     state: NoteEditorContract.UiState,
     replyingToPaddingTop: Dp,
     outlineColor: Color,
-    focusRequester: FocusRequester,
     onReplyNoteHeightChanged: (Int) -> Unit,
     onReplyToNoticeHeightChanged: (Int) -> Unit,
     eventPublisher: (UiEvent) -> Unit,
 ) {
+    val focusRequester = remember { FocusRequester() }
```
AleksandarIlic, line 494: Why here? Why not just above `NoteOutlinedTextField`? Is `awaitFrame` really necessary?

markocic: I like to have all variables and `LaunchedEffect` at the start and then keep all composable invocations in order. Just a preference. `awaitFrame` is here to ensure this crash _never_ happens again. Without that, it would be possible in some cases.

AleksandarIlic: I was thinking if we moved closer to the text field, maybe the crash will not happen even without `awaitFrame`. Thanks for explanation, let's merge it and avoid risk.

AleksandarIlic (approved):

PR #532, opened 2025-07-08
Implement playing only one video in feed with auto play

AleksandarIlic:


app/src/main/kotlin/net/primal/android/core/compose/LazyListStateUtils.kt
```diff
@@ -33,3 +39,26 @@ fun LazyListState.rememberIsItemVisible(key: Any?, fallback: Boolean): State<Boo
...
 }
+
+@Composable
+fun rememberFirstVisibleItemIndex(listState: LazyListState): MutableState<Int?> {
+    val currentlyFirstVisibleIndex = remember { mutableStateOf<Int?>(null) }
+    LaunchedEffect(listState) {
+        snapshotFlow { listState.layoutInfo }
+            .map { layoutInfo ->
+                val fullyVisibleItems = layoutInfo.visibleItemsInfo
+                    .filter {
```
markocic, line 50: Is this `filter` really necessary? `layoutInfo.visibleItemsInfo` should contain only visible items.

AleksandarIlic: Yes, we want to auto-play when full item is visible. Or at least most of the item.

markocic (approved):

PR #533, opened 2025-07-08
Fix false positive "replying to" indicator in note

mehmedalijaK:


data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/mappers/remote/PostDataEvents.kt
```diff
@@ -113,3 +129,7 @@ private fun NostrEvent.pictureNoteAsPost(): PostData {
         replyToAuthorId = null,
     )
 }
+
+private fun JsonArray.isQuoteTag(): Boolean {
```
AleksandarIlic: This should be places in the Tags file where other extension functions are for tags.

PR #534, opened 2025-07-08
Improve ensuring socket connection

AleksandarIlic:
Fixes edge case where web socket is not reconnecting and
fixes `CancellationException` proper propagation.


markocic (approved):
I was unable to reproduce this bug on the production build. Code looks good to me.

PR #546, opened 2025-07-15
Improve video autoplay logic during scroll

mehmedalijaK:


AleksandarIlic (changes requested):
Works better, but we still have edge cases to cover. Identify the edges cases (multiple notes in a feed, just one coming from the top, just one coming from the bottom, just one with lots of text in both directions, etc) and then compare against iOS app.

AleksandarIlic (approved):

PR #547, opened 2025-07-16
Fix scroll jump when returning to thread screen

mehmedalijaK:


app/src/main/kotlin/net/primal/android/thread/notes/ThreadContract.kt
```diff
@@ -16,6 +17,7 @@ interface ThreadContract {
         val fetching: Boolean = false,
         val replyToArticle: FeedArticleUi? = null,
         val error: UiError? = null,
+        val listState: LazyListState = LazyListState(),
```
AleksandarIlic, line 20: We should not remember the list state in the viewModel. We can use `rememberLazyListState()`

app/src/main/kotlin/net/primal/android/thread/notes/ThreadScreen.kt
```diff
@@ -359,9 +362,16 @@ private fun ThreadLazyColumn(
         threadListMaxHeightPx.toDp() - highlightPostHeightPx.toDp() - repliesHeightPx.values.sum().toDp()
     }
 
+    LaunchedEffect(state.highlightPostIndex) {
```
AleksandarIlic, line 365: This is a workaround. It does not fix the cause. I believe the cause is in `extraSpacing`. It gets updated too late which causes LazyColumn to have less height causing to scroll down. Not sure what would be the best way forward here for this issue.

AleksandarIlic (changes requested):

PR #548, opened 2025-07-16
Feature/migrate wallet to db

markocic:


app/src/main/kotlin/net/primal/android/wallet/dashboard/WalletDashboardViewModel.kt
```diff
@@ -116,14 +121,16 @@ class WalletDashboardViewModel @Inject constructor(
...
+    private fun fetchWalletBalance(walletId: String) =
         viewModelScope.launch {
             try {
-                walletRepository.fetchWalletBalance(userId = activeUserId)
+                walletRepository.fetchWalletBalance(walletId = walletId)
             } catch (error: SignatureException) {
                 Timber.w(error)
             } catch (error: NetworkException) {
                 Timber.w(error)
+            } catch (error: IllegalStateException) {
```
AleksandarIlic: This looks wrong.

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/repository/WalletRepositoryImpl.kt
```diff
@@ -108,21 +169,42 @@ internal class WalletRepositoryImpl(
...
-                walletDatabase.wallet().upsertWalletBalance(
-                    WalletBalance(
-                        walletId = walletInfo.localId,
-                        lastUpdatedAt = Clock.System.now().epochSeconds,
-                        balanceInBtc = response.amount,
-                        maxBalanceInBtc = response.maxAmount,
-                    ),
-                )
+            val wallet = walletDatabase.wallet().findWallet(walletId = walletId)
+                ?: error("No wallet found with the given walletId.")
```
AleksandarIlic: We should not throw error. Let's not do anything if wallet is not found. In the future we can return a result indicating if wallet was updated or not.

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/repository/WalletRepositoryImpl.kt
```diff
@@ -108,21 +169,42 @@ internal class WalletRepositoryImpl(
...
+
+            when (wallet.info.type) {
+                WalletType.PRIMAL -> {
+                    val response = primalWalletApi.getBalance(userId = walletId)
+
+                    walletDatabase.wallet().updateWalletBalance(
```
AleksandarIlic: `updateWalletBalance()` should be executed at the end for both cases. `when` should handle only the difference between wallet types.

To avoid ifs and whens: Maybe we can have some common interface for both primal wallet api and nwc api. Then construct/call the appropriate specific instance depending on the wallet type, which would return the same results which we use further.

app/src/main/kotlin/net/primal/android/wallet/dashboard/WalletDashboardViewModel.kt
```diff
@@ -39,6 +39,7 @@ import timber.log.Timber
 @HiltViewModel
 class WalletDashboardViewModel @Inject constructor(
     private val activeAccountStore: ActiveAccountStore,
+    private val walletAccountRepository: WalletAccountRepository,
```
AleksandarIlic, line 43: `userRepository` below can be just a constructor argument.

app/src/main/res/values/strings.xml
```diff
@@ -1088,7 +1088,7 @@
 
     <string name="wallet_dashboard_activate_notice_hint">You need to activate your wallet</string>
     <string name="wallet_dashboard_activate_button">Activate Wallet Now</string>
-    <string name="wallet_dashboard_enable_wallet_notice_hint">External wallet is currently enabled via Nostr Wallet Connect</string>
+    <string name="wallet_dashboard_enable_wallet_notice_hint">You don\'t have an active wallet.</string>
```
AleksandarIlic: We should never show this message to user.

data/wallet/local/src/commonMain/kotlin/net/primal/wallet/data/local/db/WalletDatabase.kt
```diff
@@ -15,13 +19,16 @@ import net.primal.wallet.data.local.dao.WalletTransactionData
...
         WalletInfo::class,
-        WalletBalance::class,
+        NostrWalletData::class,
+        PrimalWalletData::class,
+        ActiveWalletData::class,
         WalletTransactionData::class,
         WalletSettings::class,
     ],
-    version = 1,
+    version = 2,
```
AleksandarIlic: We should not increase schema while we are in a dev mode. Once we ship this internally, we can start increasing the version.

domain/wallet/src/commonMain/kotlin/net/primal/domain/wallet/WalletRepository.kt
```diff
@@ -5,7 +5,17 @@ import kotlinx.coroutines.flow.Flow
 
 interface WalletRepository {
 
-    suspend fun fetchWalletBalance(userId: String)
+    suspend fun setActiveWallet(userId: String, walletId: String)
```
AleksandarIlic: Currently active wallet functions should be in `WalletAccountRepository`. It makes sense to be under `Account`. We might need better to organize the repositories. We can not have one for sure and we should see how to divide the responsibility (especially since wallet is also handling promo codes and memberships).

domain/wallet/src/commonMain/kotlin/net/primal/domain/wallet/WalletService.kt
```diff
@@ -0,0 +1,9 @@
+package net.primal.domain.wallet
+
+import net.primal.domain.common.exception.NetworkException
+import net.primal.domain.wallet.model.WalletBalanceResult
+
+interface WalletService {
+    @Throws(NetworkException::class, kotlin.coroutines.cancellation.CancellationException::class)
```
AleksandarIlic, line 7: We should get rid off throwing exceptions and return some kind of result. Since we can not use `kotlin.Result`, maybe we can put some our own implementation in `core:utils` (all domain modules implement this module) and use it here and later everywhere?

domain/wallet/src/commonMain/kotlin/net/primal/domain/wallet/WalletRepository.kt
```diff
@@ -2,10 +2,16 @@ package net.primal.domain.wallet
...
+import net.primal.domain.common.exception.NetworkException
 
 interface WalletRepository {
 
-    suspend fun fetchWalletBalance(userId: String)
+    suspend fun upsertNostrWallet(userId: String, wallet: Wallet.NWC)
+
+    suspend fun upsertWalletSettings(walletId: String, spamThresholdAmountInSats: Long)
+
+    @Throws(NetworkException::class, kotlin.coroutines.cancellation.CancellationException::class)
```
AleksandarIlic, line 13: Let's use result here as well; and in other wallet repository.

AleksandarIlic (approved):
All looks good. Comments above can be fixed in this or later PRs. It's up to you. And thank you for great work!

PR #554, opened 2025-07-18
Add support for immersive mode for videos

mehmedalijaK:


markocic (changes requested):
Works great. One thing I have noticed is that entering and exiting immersive mode causes the video to move up and down. We should try to adjust screen so the video stays in place.

AleksandarIlic (changes requested):
We should sync immersive mode with video controls, similarly like iOS app is doing.

AleksandarIlic (approved):

PR #555, opened 2025-07-18
Feature/kmp wallet transactions

markocic:


data/wallet/remote-nwc/build.gradle.kts
```diff
@@ -40,6 +40,7 @@ kotlin {
                 implementation(project(":core:networking-http"))
                 implementation(project(":core:networking-primal"))
                 implementation(project(":domain:nostr"))
+                implementation(project(":domain:wallet"))
```
AleksandarIlic: Why do we need dependency on wallet domain here? We should not do that, data-remote should not have dependency on domain.

app/src/main/kotlin/net/primal/android/wallet/dashboard/WalletDashboardViewModel.kt
```diff
@@ -83,11 +81,26 @@ class WalletDashboardViewModel @Inject constructor(
             }
         }
 
+    private fun subscribeToActiveWalletId() =
```
AleksandarIlic, line 85: Can we call the other subscribe function `subscribeToActiveWalletData()`?

domain/wallet/src/commonMain/kotlin/net/primal/domain/transactions/Transaction.kt
```diff
@@ -0,0 +1,102 @@
+package net.primal.domain.transactions
+
+import net.primal.domain.profile.ProfileData
+import net.primal.domain.wallet.SubWallet
+import net.primal.domain.wallet.TxState
+import net.primal.domain.wallet.TxType
+import net.primal.domain.wallet.WalletType
+
+sealed class Transaction(
```
AleksandarIlic, line 8: I think we should reconsider our domain models. Instead of grouping by source, we should organize into Lightning, OnChain, Zap, etc.

AleksandarIlic (changes requested):

domain/wallet/src/commonMain/kotlin/net/primal/domain/wallet/WalletService.kt
```diff
@@ -1,9 +1,12 @@
...
-import net.primal.domain.common.exception.NetworkException
+import net.primal.core.utils.Result
+import net.primal.domain.transactions.Transaction
+import net.primal.domain.wallet.model.TransactionsRequest
 import net.primal.domain.wallet.model.WalletBalanceResult
 
 interface WalletService {
-    @Throws(NetworkException::class, kotlin.coroutines.cancellation.CancellationException::class)
-    suspend fun fetchWalletBalance(wallet: Wallet): WalletBalanceResult
+    suspend fun fetchWalletBalance(wallet: Wallet): Result<WalletBalanceResult>
```
AleksandarIlic: Not sure if we should put `WalletService` in domain. This might be the impl. details.

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/repository/mappers/local/WalletTransaction.kt
```diff
@@ -1,20 +1,46 @@
...
+        otherUserId = this.primal?.otherUserId,
+        otherLightningAddress = this.primal?.otherLightningAddress,
+        otherUserProfile = otherProfile,
+    )
 
-            Transaction.Primal(
+private fun WalletTransactionPO.fromPrimalTxToDomain(otherProfile: ProfileData?): TransactionDO =
+    when {
+        this.primal?.isZap == true -> {
+            TransactionDO.Zap(
```
AleksandarIlic: Maybe we should have additional specific extension functions like: `WalletTransactionPO.asZapTransactionDO()` so we don't end up with couple of large constructors in a single function. And I guess this can be then reused across the module.

PR #559, opened 2025-07-22
Improve article images rendering

mehmedalijaK:


app/src/main/kotlin/net/primal/android/thread/articles/details/ArticleDetailsScreen.kt
```diff
@@ -696,29 +696,36 @@ private fun ArticleContentWithComments(
...
-                                }
-                            },
-                        model = part.imageUrl,
-                        contentScale = ContentScale.FillWidth,
-                        contentDescription = null,
-                    )
+                            .background(color = AppTheme.colorScheme.surfaceVariant),
+                        contentAlignment = Alignment.Center,
+                    ) {
+                        SubcomposeAsyncImage(
```
AleksandarIlic: Should we use `PrimalAsyncImage` here? @markocic

markocic: Technically, yes. This image is in a `LazyColumn` but given the scenario here - performance gain would be minimal.

AleksandarIlic: Let's use `PrimalAsyncImage` just to be in sync with the rest of the app.

AleksandarIlic (approved):

PR #560, opened 2025-07-22
Fix notifications sort behavior when toggling switches

mehmedalijaK:


app/src/main/kotlin/net/primal/android/settings/notifications/NotificationsSettingsScreen.kt
```diff
@@ -173,6 +173,8 @@ fun <T : NotificationSettingsType> NotificationsSettingsBlock(
     notifications: List<NotificationSwitchUi<T>>,
     eventPublisher: (NotificationsSettingsContract.UiEvent) -> Unit,
 ) {
+    val orderedNotifications = notifications.sortedBy { it.settingsType.order }
```
AleksandarIlic: This should be in `remember`.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #564, opened 2025-07-28
Implement live streaming avatar indicator on profile details screen

mehmedalijaK:


data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/streams/StreamData.kt
```diff
@@ -0,0 +1,24 @@
...
+    val dTag: String,
+    val title: String?,
+    val summary: String?,
+    val imageUrl: String?,
+    val hashtags: List<String>,
+    val streamingUrl: String?,
+    val recordingUrl: String?,
+    val startsAt: Long?,
+    val endsAt: Long?,
+    val status: String,
```
markocic: Could this be an enum?

data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/streams/StreamData.kt
```diff
@@ -0,0 +1,24 @@
...
+    @PrimaryKey
+    val authorId: String,
+    val dTag: String,
+    val title: String?,
+    val summary: String?,
+    val imageUrl: String?,
+    val hashtags: List<String>,
+    val streamingUrl: String?,
+    val recordingUrl: String?,
+    val startsAt: Long?,
```
markocic, line 19: Is this nullable? If I remember correctly, all live events must provide this.

data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/streams/StreamData.kt
```diff
@@ -0,0 +1,24 @@
+package net.primal.data.local.dao.streams
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity
+data class StreamData(
+    @PrimaryKey
+    val authorId: String,
```
markocic, line 11: Are we keeping historic data of past streams? I know the NIP provides recording of the stream after it ends. If so, primary key for this entity should be a composite key of authorId and dTag.

Might be good to consider using aTag since this is a replaceable event.

data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/streams/StreamData.kt
mehmedalijaK: We're currently working on kind `30311`, not `30313`.
For `30311`, the only required tag is the `["d", "<unique identifier>"]` tag — all other tags, including the one you're referring to, are optional per the draft spec.

app/src/main/kotlin/net/primal/android/core/compose/UniversalAvatarThumbnail.kt
```diff
@@ -66,26 +67,37 @@ fun UniversalAvatarThumbnail(
...
-        sourceUrl = avatarCdnImage?.sourceUrl,
-        blossoms = avatarBlossoms,
-        hasOuterBorder = hasBorder && avatarSize > 0.dp,
-        hasInnerBorder = hasLegendBorder && avatarSize > 0.dp && hasInnerBorderOverride,
-        borderBrush = borderBrush ?: Brush.linearGradient(
-            colors = listOf(
-                fallbackBorderColor,
-                fallbackBorderColor,
+    Box(
+        modifier = modifier.size(avatarSize + totalBorderSize * 2),
```
AleksandarIlic, line 71: Are you sure this Box is not causing any layout changes to the avatar compared to previous version? We are using avatars everywhere.

data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/streams/StreamData.kt
AleksandarIlic: We should not use `authorId`, instead we could use the same approach as with articles. It's the same data type, just different kind.

data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/streams/StreamDataDao.kt
```diff
@@ -0,0 +1,19 @@
+package net.primal.data.local.dao.streams
+
+import androidx.room.Dao
+import androidx.room.Query
+import androidx.room.Upsert
+import kotlinx.coroutines.flow.Flow
+
+@Dao
+interface StreamDataDao {
```
AleksandarIlic, line 10: I think we should take same approach as with articles.

data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/streams/StreamDataDao.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    @Upsert
+    suspend fun upsertStreamData(data: StreamData)
+
+    @Query("SELECT * FROM StreamData WHERE authorId = :authorId")
+    fun observeStream(authorId: String): Flow<StreamData?>
```
AleksandarIlic: You should have a DAO function which return all live streams of the give user and then on the repository or viewModel level you take just the first.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/profile/ProfileRepositoryImpl.kt
```diff
@@ -129,6 +131,12 @@ class ProfileRepositoryImpl(
                 if (profileStats != null) {
                     database.profileStats().upsert(data = profileStats)
                 }
+
+                if (streamData != null) {
+                    database.streams().upsertStreamData(data = streamData)
+                } else {
+                    database.streams().deleteStream(authorId = profileId)
```
AleksandarIlic: Delete is not necessary once you implement multiple streams support. We are only adding to the database what we receive and that's it. Just like with articles or any other data.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -0,0 +1,19 @@
...
+import net.primal.data.local.db.PrimalDatabase
+import net.primal.data.repository.mappers.local.asStreamDO
+import net.primal.domain.streams.Stream
+import net.primal.domain.streams.StreamRepository
+
+class StreamRepositoryImpl(
+    private val database: PrimalDatabase,
+) : StreamRepository {
+
+    override fun observeStream(authorId: String): Flow<Stream?> {
```
AleksandarIlic: This could be `observeLatestStream` which would get the latest stream id and then starts observing that stream id. This is important in case some user adds one more live stream - we want to keep observing the existing stream and not to switch automatically to the new latest stream.

PR #568, opened 2025-07-30
Implement video player initial screen

mehmedalijaK:


app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsViewModel.kt
```diff
@@ -344,12 +347,24 @@ class ProfileDetailsViewModel @Inject constructor(
...
                     setState {
-                        copy(isLive = streamData?.isLive() == true)
+                        copy(
+                            isLive = isLive,
+                            liveStreamNaddr = if (isLive) {
+                                Naddr(
+                                    kind = NostrEventKind.LiveActivity.value,
+                                    userId = streamData.authorId,
+                                    identifier = streamData.dTag,
+                                ).toNaddrString()
```
AleksandarIlic: `toNaddrString()` function on the `Stream` would be nice to have.

app/src/main/kotlin/net/primal/android/stream/StreamInfoSection.kt
```diff
@@ -0,0 +1,92 @@
...
+    authorProfile: ProfileDetailsUi,
+    viewers: Int,
+    startedAt: Long?,
+) {
+    val numberFormat = remember { NumberFormat.getNumberInstance() }
+
+    Column(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = 16.dp, vertical = 15.dp),
```
AleksandarIlic: Is this 1.dp really making a difference?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+            val naddr = parseAndResolveNaddr()
+            if (naddr != null) {
+                observeStream(naddr = naddr)
+            } else {
+                Timber.w("Unable to resolve naddr.")
+                setState { copy(loading = false) }
+            }
+        }
+
+    private suspend fun parseAndResolveNaddr(): Naddr? {
```
AleksandarIlic: This could be done `inline` to avoid having two resolving functions.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+                .filterNotNull()
+                .collect { stream ->
+                    val streamingUrl = stream.streamingUrl
+                    if (streamingUrl == null) {
+                        setState { copy(loading = false) }
+                        return@collect
+                    }
+
+                    val authorId = stream.authorId
+                    profileRepository.observeProfileData(profileId = authorId).collect { profileData ->
```
AleksandarIlic: This is not a good approach. This is going to block further events from `observeStream`. I think you already have `authorId` from `naddr`, can you observe in parallel? One observer for stream and another for stream author profile data?

app/src/main/kotlin/net/primal/android/stream/LiveStreamScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+        }
+        exoPlayer.addListener(listener)
+
+        onDispose {
+            exoPlayer.removeListener(listener)
+            exoPlayer.release()
+        }
+    }
+
+    LaunchedEffect(state.isPlaying, state.isLive, state.isSeeking) {
```
AleksandarIlic: Why do we need this Effect? Are you sure you really need to restart this effect every time this values change since the code is in while (true)?

app/src/main/kotlin/net/primal/android/stream/LiveStreamPlayerControls.kt
```diff
@@ -0,0 +1,355 @@
...
+import net.primal.android.theme.AppTheme
+
+private const val SECONDS_IN_MINUTE = 60
+
+@Composable
+fun LiveStreamPlayerControls(
+    modifier: Modifier = Modifier,
+    isVisible: Boolean,
+    state: LiveStreamContract.UiState,
+    eventPublisher: (LiveStreamContract.UiEvent) -> Unit,
```
AleksandarIlic: I think you are using this to emit only single event type. Can you create a specific callback instead?

app/src/main/kotlin/net/primal/android/stream/LiveStreamPlayerControls.kt
```diff
@@ -0,0 +1,355 @@
...
+import net.primal.android.core.compose.icons.primaliconpack.More
+import net.primal.android.theme.AppTheme
+
+private const val SECONDS_IN_MINUTE = 60
+
+@Composable
+fun LiveStreamPlayerControls(
+    modifier: Modifier = Modifier,
+    isVisible: Boolean,
+    state: LiveStreamContract.UiState,
```
AleksandarIlic: Instead of passing around the screen state, you could create a stream data class which you can use in the state. And then instead of the state you can pass around that particular stream data.

app/src/main/kotlin/net/primal/android/stream/LiveStreamPlayer.kt
```diff
@@ -0,0 +1,97 @@
...
+private const val VIDEO_ASPECT_RATIO_WIDTH = 16f
+private const val VIDEO_ASPECT_RATIO_HEIGHT = 9f
+
+@androidx.annotation.OptIn(UnstableApi::class)
+@Composable
+fun LiveStreamPlayer(
+    exoPlayer: ExoPlayer,
+    streamUrl: String,
+    state: LiveStreamContract.UiState,
+    eventPublisher: (LiveStreamContract.UiEvent) -> Unit,
```
AleksandarIlic: Same comments here for `state` and `eventPublisher` as from player controls.

app/src/main/kotlin/net/primal/android/stream/LiveStreamScreen.kt
```diff
@@ -0,0 +1,216 @@
...
+@Composable
+private fun LiveStreamContent(
+    state: LiveStreamContract.UiState,
+    exoPlayer: ExoPlayer,
+    eventPublisher: (LiveStreamContract.UiEvent) -> Unit,
+    paddingValues: PaddingValues,
+    onClose: () -> Unit,
+) {
+    if (state.loading) {
+        PrimalLoadingSpinner()
```
AleksandarIlic, line 151: Do we need loading spinner here and as well in the player? Or is this for the initial data loading and player is for playing?

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -87,34 +109,94 @@ class LiveStreamViewModel @Inject constructor(
...
-                        setState { copy(loading = false) }
-                        return@collect
-                    }
+            val authorId = naddr.userId
+            combine(
+                streamRepository.observeStream(aTag = naddr.asATagValue()).filterNotNull(),
+                profileRepository.observeProfileData(profileId = authorId),
+                profileRepository.observeProfileStats(profileId = authorId),
+                activeAccountStore.activeUserAccount,
+            ) { stream, profileData, profileStats, activeUserAccount ->
```
AleksandarIlic: Is is really required to combine them all? Can we have them all separate? @markocic what do you think?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -87,34 +109,94 @@ class LiveStreamViewModel @Inject constructor(
...
+                    shouldApproveProfileAction = null,
+                    isFollowed = actions.firstOrNull() is ProfileFollowsHandler.Action.Follow,
+                )
+            }
+            profileFollowsHandler.forceUpdateList(actions = actions)
+        }
+
+    private fun observeFollowsResults() =
+        viewModelScope.launch {
+            profileFollowsHandler.observeResults().collect {
```
AleksandarIlic, line 207: @markocic Can you review the following code in this PR?

markocic: Follows logic seems to closely mirror one in `ProfileDetailsViewModel`. This looks great to me!

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
markocic: This does make the code a lot simpler to read but there are downsides. Using `combine` the stream won't be emitted until every stream emits one value. Notable downside is that change in one stream later on will trigger the whole `transform` block to run again. We don't have any expensive functions running in this `transform` block so this _could_ be negligible. Might be best not to introduce unnecessary complexity as in future we should pay attention to things inside this `transform` block.

PR #571, opened 2025-08-01
Feature/kmp wallet deprecate user account

markocic:


app/src/main/kotlin/net/primal/android/wallet/activation/WalletActivationViewModel.kt
```diff
@@ -167,55 +165,45 @@ class WalletActivationViewModel @Inject constructor(
...
-                try {
-                    promoCode?.let {
-                        walletAccountRepository.redeemPromoCode(
-                            userId = activeAccountStore.activeUserId(),
-                            code = promoCode,
+            walletAccountRepository.activateWallet(userId, code)
+                .alsoCatching { response ->
+                    walletAccountRepository.fetchWalletAccountInfo(userId)
+                        .onSuccess { walletAccountRepository.setActiveWallet(userId = userId, walletId = userId) }
+                }.onSuccess { response ->
```
AleksandarIlic: Can we here link them somehow so that we end up with just one `onSuccess` and one `onError`?

app/src/main/kotlin/net/primal/android/wallet/activation/WalletActivationViewModel.kt
```diff
@@ -167,55 +165,45 @@ class WalletActivationViewModel @Inject constructor(
...
-                    }
-                }
+            val userId = activeAccountStore.activeUserId()
 
-                try {
-                    promoCode?.let {
-                        walletAccountRepository.redeemPromoCode(
-                            userId = activeAccountStore.activeUserId(),
-                            code = promoCode,
+            walletAccountRepository.activateWallet(userId, code)
```
AleksandarIlic, line 170: Can we extract this complex logic into PrimalWalletActiviation use case or handler?

app/src/main/kotlin/net/primal/android/settings/wallet/nwc/scan/NwcQrCodeScannerViewModel.kt
```diff
@@ -63,15 +67,36 @@ class NwcQrCodeScannerViewModel @Inject constructor(
...
+                        privateKey = nostrWalletConnect.keypair.privateKey,
+                        pubKey = nostrWalletConnect.keypair.pubkey,
+                    ),
+                ),
+            )
+
+            walletAccountRepository.setActiveWallet(
+                userId = activeAccountStore.activeUserId(),
+                walletId = walletId,
+            )
```
AleksandarIlic: Is the above code block the same or very similar as in `WalletSettingsViewModel`? If yes, we could have it in a use case or handler and reuse that in both places.

app/src/main/kotlin/net/primal/android/wallet/activation/WalletActivationViewModel.kt
markocic: We would have trouble with this flow being dependent on `UserRepository`. I believe we could overcome this. Caller would have to do that manually.

One more issue that could arise is due to the complex error handling here. Same exceptions mean different UI messages depending on the operation that failed. We could encapsulate them in our own exceptions and return them as such allowing us to keep the same logic after migration.

app/src/main/kotlin/net/primal/android/wallet/activation/WalletActivationViewModel.kt
markocic: I don't think we can. Failure of some of these operation are tolerable while others are not. We have to be aware of exactly which operation has failed as to handle the error correctly.

- If the failure occurs on the main flow (`activateWallet` or `fetchWalletAccountInfo`), We are going to terminate and set `error`.
- If the `redeemPromoCode` fails, we shouldn't terminate as user can always redeem code once wallet has been activated. Similar thing applies to `setLightningAddress` but we are just silently failing that one compared to promo code failure in which we notify the user.

There are small exceptions to each case which is why we ended up with this solution. We could rethink some of these cases but I am not sure if we should. This flow happens only once and in this one place. It's low impact enough for us to keep it like this for now.

Let me know your thoughts on this and we can proceed from there.

app/src/main/kotlin/net/primal/android/wallet/activation/WalletActivationViewModel.kt
AleksandarIlic: Oh, I see. We can migrate this once we kmpify `UserRepository`.

app/src/main/kotlin/net/primal/android/wallet/activation/WalletActivationViewModel.kt
AleksandarIlic: Let's then go with something like
```kotlin
          walletAccountRepository.activateWallet(userId, code)
                .alsoCatching { walletAccountRepository.fetchWalletAccountInfo(userId) }
                .onSuccess { response -> ...
```

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/usecase/ConnectNwcUseCaseImpl.kt
```diff
@@ -0,0 +1,47 @@
...
+import net.primal.core.utils.asSha256Hash
+import net.primal.core.utils.runCatching
+import net.primal.domain.account.WalletAccountRepository
+import net.primal.domain.usecase.ConnectNwcUseCase
+import net.primal.domain.wallet.NostrWalletKeypair
+import net.primal.domain.wallet.Wallet
+import net.primal.domain.wallet.WalletRepository
+import net.primal.wallet.data.parser.parseNWCUrl
+
+class ConnectNwcUseCaseImpl(
```
AleksandarIlic: Let's put this as concrete class directly in the `wallet-domain` module instead of having impl. in data layer.

domain/wallet/src/commonMain/kotlin/net/primal/domain/usecase/ConnectNwcUseCase.kt
```diff
@@ -0,0 +1,7 @@
+package net.primal.domain.usecase
+
+import net.primal.core.utils.Result
+
+interface ConnectNwcUseCase {
+    suspend fun connectNWCWallet(userId: String, nwcUrl: String): Result<Unit>
```
AleksandarIlic: Name should be `invoke` or something like that. Use case will have only one function.

AleksandarIlic (changes requested):

domain/wallet/src/commonMain/kotlin/net/primal/domain/usecase/factory/WalletUseCaseFactory.kt
```diff
@@ -1,16 +1,15 @@
-package net.primal.wallet.data.usecase.factory
+package net.primal.domain.usecase.factory
 
 import net.primal.domain.account.WalletAccountRepository
 import net.primal.domain.usecase.ConnectNwcUseCase
 import net.primal.domain.wallet.WalletRepository
-import net.primal.wallet.data.usecase.ConnectNwcUseCaseImpl
 
 object WalletUseCaseFactory {
```
AleksandarIlic: Let's remove factory class. We can use constructor directly now.

PR #573, opened 2025-08-01
Feature/live stream zap pills

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -59,6 +73,47 @@ class LiveStreamViewModel @Inject constructor(
         observeFollowsResults()
     }
 
+    override fun onCleared() {
+        liveFeedMonitor.stopMonitor(viewModelScope)
+    }
+
+    private fun startLiveStreamMonitor(naddr: Naddr) {
+        liveFeedMonitor.startMonitor(
```
AleksandarIlic: This entire code should not be happening in the view model. We should just say something like `liveStreamingMintor.start(naddr)` and the magic happens under the hood. All the necessary data is then available in the ViewModel by accessing local database.

app/src/main/kotlin/net/primal/android/stream/subscription/LiveFeedMonitor.kt
```diff
@@ -0,0 +1,81 @@
...
+    @PrimalCacheApiClient private val primalApiClient: PrimalApiClient,
+) {
+    private var monitorSubscription: PrimalSocketSubscription<NostrEvent>? = null
+    private val monitorMutex = Mutex()
+
+    fun startMonitor(
+        scope: CoroutineScope,
+        creatorPubkey: String,
+        dTag: String,
+        userPubkey: String,
```
AleksandarIlic: We should pass `scope`, `naddr` and `userId`. Everything else is not necessary.

domain/primal/src/commonMain/kotlin/net/primal/domain/streams/StreamRepository.kt
```diff
@@ -1,10 +1,13 @@
...
 import kotlinx.coroutines.flow.Flow
+import net.primal.domain.nostr.NostrEvent
 
 interface StreamRepository {
 
     suspend fun findLatestLiveStreamATag(authorId: String): String?
 
     fun observeStream(aTag: String): Flow<Stream?>
+
+    suspend fun saveZap(zapEvent: NostrEvent, zappedEventATag: String)
```
AleksandarIlic: We should not do this. Monitor should be part of the `caching-data` module having access directly to the database.

app/src/main/kotlin/net/primal/android/stream/subscription/LiveFeedMonitor.kt
```diff
@@ -0,0 +1,81 @@
...
+import kotlinx.coroutines.sync.withLock
+import net.primal.android.networking.di.PrimalCacheApiClient
+import net.primal.core.networking.primal.PrimalApiClient
+import net.primal.core.networking.primal.PrimalCacheFilter
+import net.primal.core.networking.primal.PrimalSocketSubscription
+import net.primal.core.utils.serialization.encodeToJsonString
+import net.primal.domain.nostr.NostrEvent
+import net.primal.domain.nostr.NostrEventKind
+
+class LiveFeedMonitor @Inject constructor(
```
AleksandarIlic: This should belong to the `data/caching` modules. Talking directly with our caching service should be part of `caching:remote` and the `app` module is going to talk only with the `caching:repostiory` module. We probably need both Monitor (in `caching:rmeote`) and something else (in `caching:repository`) which is going to accept incoming events and process them in a batches.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -19,9 +39,77 @@ class StreamRepositoryImpl(
...
+
+    override suspend fun startLiveStreamSubscription(naddr: Naddr, userId: String) =
+        coroutineScope {
+            streamMonitor.subscribe(
+                streamingNaddr = naddr,
+                userId = userId,
+            ).collect { response ->
+                response.zaps.forEach {
+                    processIncomingZap(zapEvent = it)
+                }
```
AleksandarIlic: The entire processing logic needs to be changed to include the new architecture. We need to process by `LiveStreamResponse` which can include multiple nostr or primal events.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -19,9 +39,77 @@ class StreamRepositoryImpl(
...
+        withContext(dispatcherProvider.io()) {
+            try {
+                val zapRequest = zapEvent.extractZapRequestOrNull()
+                val zapperPubkey = zapRequest?.pubKey
+                if (zapperPubkey == null) {
+                    Napier.w("Unable to extract zapper pubkey from zap event.")
+                    return@withContext
+                }
+
+                val localProfile = database.profiles().findProfileData(profileId = zapperPubkey)
```
AleksandarIlic: Additional api calls should be async and must not block the collector. Collector can be blocked only by processing the current response and writing it to database, any additional work must happen asynchronously.

AleksandarIlic (changes requested):
Please address the final comments to adjust to the new architecture.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -1,27 +1,121 @@
...
+            streamingNaddr = naddr,
+            userId = userId,
+        ).collect { response ->
+            processLiveStreamResponse(response = response)
+        }
+    }
+
+    private suspend fun processLiveStreamResponse(response: LiveFeedResponse) {
+        val zapEvents = response.zaps
+        if (zapEvents.isEmpty()) return
```
AleksandarIlic: We don't want to return if there are no zaps since we are going to process here multiple events in the future, not only zaps. It should work perfectly with or without zaps. The body structure of this function should be similar like the body in `persistToDatabaseAsTransaction`. Something like this:
1) You gather all the necessary events;
2) You process and extract whatever is necessary;
3) Once you have everything ready, you open database transaction and make database changes;
4) If there are some additional async work left to be done this is the place where you trigger that async work;

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -1,27 +1,121 @@
...
+
+        val zapRequests = zapEvents.mapNotNull { it.extractZapRequestOrNull() }
+        val zapperPubkeys = zapRequests.map { it.pubKey }.toSet()
+
+        if (zapperPubkeys.isNotEmpty()) {
+            val localProfiles = withContext(dispatcherProvider.io()) {
+                database.profiles().findProfileData(profileIds = zapperPubkeys.toList())
+            }
+            withContext(dispatcherProvider.io()) {
+                saveZaps(zapEvents = zapEvents, profiles = localProfiles)
```
AleksandarIlic: Add `saveZaps` logic inline here and remove this function. All inserts to database should happen within transaction and this is why we want like this - later we are going to add to that transaction other inserts.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -1,27 +1,121 @@
...
+                val cdnResources = response.cdnResources.flatMapNotNullAsCdnResource()
+                val blossomServers = response.blossomServers.mapAsMapPubkeyToListOfBlossomServers()
+                val profiles = response.metadataEvents.mapAsProfileDataPO(
+                    cdnResources = cdnResources,
+                    primalUserNames = primalUserNames,
+                    primalPremiumInfo = primalPremiumInfo,
+                    primalLegendProfiles = primalLegendProfiles,
+                    blossomServers = blossomServers,
+                )
+                database.profiles().insertOrUpdateAll(data = profiles)
```
AleksandarIlic: Code like this belongs to `ProfileRepository`. Can you check if already have function like this maybe? If yes, then this means that you should use it. We can see later if we need to extract the entire processing logic into separate Handler or UseCase class so we don't end up with one repository depending on another.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -1,27 +1,121 @@
...
+import net.primal.domain.nostr.Naddr
+import net.primal.domain.nostr.NostrEvent
 import net.primal.domain.streams.Stream
 import net.primal.domain.streams.StreamRepository
 
 class StreamRepositoryImpl(
+    private val dispatcherProvider: DispatcherProvider,
     private val database: PrimalDatabase,
+    private val usersApi: UsersApi,
+    private val streamMonitor: LiveStreamApi,
```
AleksandarIlic: Please rename it to `liveStreamApi`.

AleksandarIlic (approved):
We should avoid repository depending on another repository once we establish a better pattern and example afterwards in a separate PR. Great job!

PR #583, opened 2025-08-08
Improve chat system and add overflow menu

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -386,6 +427,150 @@ class LiveStreamViewModel @Inject constructor(
...
+            } catch (error: NostrPublishException) {
+                Timber.w(error)
+            }
+        }
+
+    private fun requestDeleteStream() =
+        viewModelScope.launch {
+            val streamInfo = state.value.streamInfo ?: return@launch
+            val activeUserId = state.value.activeUserId ?: return@launch
+            if (streamInfo.authorId != activeUserId) return@launch
```
AleksandarIlic: Can you do all this checks under 1 if and 1 early return?

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamMenu.kt
```diff
@@ -0,0 +1,237 @@
...
+import net.primal.android.core.utils.copyText
+import net.primal.android.core.utils.systemShareText
+import net.primal.android.profile.report.ReportUserDialog
+import net.primal.android.theme.AppTheme
+import net.primal.domain.nostr.Naddr
+import net.primal.domain.nostr.Nip19TLV.toNaddrString
+import net.primal.domain.nostr.ReportType
+import net.primal.domain.nostr.utils.withNostrPrefix
+
+private fun resolvePrimalStreamLink(naddr: String): String = "https://primal.net/a/$naddr"
```
AleksandarIlic: I think we have some utils class where all functions for this are located. Can you find it and move there?

PR #587, opened 2025-08-11
Fix avatar live indicator in feeds

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/compose/UniversalAvatarThumbnail.kt
```diff
@@ -50,6 +50,7 @@ fun UniversalAvatarThumbnail(
     hasInnerBorderOverride: Boolean = true,
     forceAnimationIfAvailable: Boolean = false,
     isLive: Boolean = false,
+    minLiveContainerWidth: Dp? = null,
```
AleksandarIlic: This field is always `null` or 50.dp. Is this something that could be done with some `Boolean`?

app/src/main/kotlin/net/primal/android/core/compose/LiveChip.kt
```diff
@@ -13,49 +13,113 @@ import androidx.compose.ui.Alignment
...
+    const val FONT_SIZE_SCALE_FACTOR = 0.140f
+    const val LETTER_SPACING_SCALE_FACTOR = 0.011f
+}
+
 @Composable
-fun LiveChip(modifier: Modifier = Modifier) {
+fun LiveChip(
+    modifier: Modifier = Modifier,
+    containerWidth: Dp,
+    minContainerWidth: Dp? = null,
```
AleksandarIlic: Both `containerWidth` and `minContainerWidth` looks strange. You could probably calculated the `calculationWidth` directly at the call place and just have one argument.

app/src/main/kotlin/net/primal/android/stream/ui/StreamInfoSection.kt
```diff
@@ -189,7 +189,7 @@ fun StreamLiveIndicator(modifier: Modifier = Modifier, isLive: Boolean) {
                 ),
         )
         Text(
-            text = stringResource(id = R.string.live_stream_chip_title),
+            text = stringResource(id = R.string.live_stream_live_indicator),
```
AleksandarIlic, line 192: Keep it `Live` in `strings.xml` and call `.uppercase()` here. This is important for translations, if someone translate differently we are here making sure that is always in upper case.

AleksandarIlic (changes requested):
UX looks great. Just a few minor improvements.

PR #589, opened 2025-08-12
Add support for user mentions in chat messages

mehmedalijaK:


AleksandarIlic (changes requested):
Is there a way to encapsulate the user mentioning logic into specific component which could be reusable across screens instead of just adding the same code here? Maybe it could have it's own ViewModel with it's own state which we just plug-in into desired screens?

PR #592, opened 2025-08-13
Feature/kmp wallet db encryption

markocic:


AleksandarIlic (dismissed):

app/src/main/kotlin/net/primal/android/PrimalApp.kt
```diff
@@ -35,7 +35,7 @@ class PrimalApp : Application() {
         super.onCreate()
         AppConfigInitializer.init(this@PrimalApp)
         PrimalRepositoryFactory.init(this@PrimalApp)
-        WalletRepositoryFactory.init(this@PrimalApp)
+        WalletRepositoryFactory.init(this@PrimalApp, !BuildConfig.DEBUG)
```
AleksandarIlic: We should use named arguments here. It's unclear what `!BuildConfig.DEBUG` is for.

data/wallet/repository/src/iosMain/kotlin/net/primal/wallet/data/repository/factory/IosRepositoryFactory.kt
```diff
@@ -11,5 +11,9 @@ object IosRepositoryFactory : RepositoryFactory() {
         LocalDatabaseFactory.createDatabase<WalletDatabase>(databaseName = "wallet_database.db")
     }
 
+    fun init(enableEncryption: Boolean) {
```
AleksandarIlic: It's unclear to what encryptions refers. Can we use something like `enableDbEncryption` both here and in Android factory?

AleksandarIlic (approved):

PR #598, opened 2025-08-18
Implement new stream description bottom sheet

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/overlay/LiveStreamOverlay.kt
```diff
@@ -74,16 +77,47 @@ private fun LiveStreamOverlay(
         is StreamMode.Expanded -> {
             viewModel.setEvent(LiveStreamContract.UiEvent.StartStream(mode.naddr))
 
-            LiveStreamScreen(
-                viewModel = viewModel,
-                state = uiState.value,
-                exoPlayer = exoPlayer,
+            val callbacks = LiveStreamContract.ScreenCallbacks(
```
AleksandarIlic: This is great change. It's much better to have screen callbacks than using note callbacks. I would just like to extract this boilerplate code into a separate function (just like we had with note callbacks) so we can keep this function clean.

app/src/main/kotlin/net/primal/android/stream/ui/StreamInfoSection.kt
```diff
@@ -106,7 +106,7 @@ fun StreamInfoSection(
 }
 
 @Composable
-private fun StreamMetaData(
+fun StreamMetaData(
```
AleksandarIlic: If this is reusable across the different composable components then let's put it in it's own file.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -268,21 +273,71 @@ private fun LiveStreamScreen(
...
+                        onZap = onZapClick,
+                        onEditProfileClick = callbacks.onEditProfileClick,
+                        onMessageClick = callbacks.onMessageClick,
+                        onDrawerQrCodeClick = callbacks.onDrawerQrCodeClick,
+                        onHashtagClick = callbacks.onHashtagClick,
+                    )
+                }
+            }
+        }
+        is ActiveBottomSheet.None -> {
```
AleksandarIlic: Let's use `Unit` instead of empty brackets when we must.

PR #600, opened 2025-08-18
Improve LiveStreamViewModel to be scoped to naddr

AleksandarIlic:
This PR also:
- Removes bookmarks from live streams;
- Introduces ZapHost solution for zapping boilerplate;

### DIFF
--- a/app/src/main/kotlin/net/primal/android/stream/player/StreamMode.kt
+++ b/app/src/main/kotlin/net/primal/android/stream/player/StreamMode.kt
@@ -2,7 +2,16 @@ package net.primal.android.stream.player
 
 sealed interface StreamMode {
     object Closed : StreamMode
-    data class Hidden(val naddr: String, val modeToRestore: StreamMode) : StreamMode
+    data class Hidden(val modeToRestore: StreamMode) : StreamMode
     data class Minimized(val naddr: String) : StreamMode
     data class Expanded(val naddr: String) : StreamMode
+
+    fun resolveNaddr(): String? {
+        return when (val mode = this) {
+            is Expanded -> mode.naddr
+            is Minimized -> mode.naddr
+            is Hidden -> mode.modeToRestore.resolveNaddr()
+            else -> null
+        }
+    }
 }
--- a/app/src/main/kotlin/net/primal/android/stream/player/StreamState.kt
+++ b/app/src/main/kotlin/net/primal/android/stream/player/StreamState.kt
@@ -42,13 +42,9 @@ class StreamState internal constructor(
     }
 
     fun hide() {
-        val naddr = when (_mode) {
-            is StreamMode.Minimized -> (_mode as StreamMode.Minimized).naddr
-            is StreamMode.Expanded -> (_mode as StreamMode.Expanded).naddr
-            else -> null
-        }
-        if (naddr != null) {
-            _mode = StreamMode.Hidden(naddr = naddr, modeToRestore = _mode)
+        val current = _mode
+        if (current !is StreamMode.Hidden) {
+            _mode = StreamMode.Hidden(modeToRestore = current)
         }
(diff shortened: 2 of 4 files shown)


app/src/main/kotlin/net/primal/android/notes/feed/zaps/ZapHost.kt
```diff
@@ -0,0 +1,81 @@
...
+            showZapOptions = true
+        } else {
+            showCantZapWarning = true
+        }
+    }
+}
+
+@Composable
+fun rememberZapHostState(zappingState: ZappingState, receiverName: String? = null): ZapHostState =
+    remember(zappingState, receiverName) {
```
markocic, line 76: Not sure if we should use `rememberSaveable` here. Configuration changes do kill bottom sheet but zapping behavior is kinda buggy right now so I can't confirm anything beyond the obvious.

app/src/main/kotlin/net/primal/android/stream/player/StreamMode.kt
```diff
@@ -2,7 +2,7 @@ package net.primal.android.stream.player
 
 sealed interface StreamMode {
     object Closed : StreamMode
-    data class Hidden(val modeToRestore: StreamMode) : StreamMode
+    data class Hidden(val naddr: String, val modeToRestore: StreamMode) : StreamMode
```
markocic: Not sure if we should keep `naddr` in `Hidden` state. I don't think `Hidden` state should know that. We can resolve `naddr` recursively through `modeToRestore`.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -118,99 +111,30 @@ private val ZapMessageProfileHandleColor: Color
...
-        ApproveBookmarkAlertDialog(
-            onBookmarkConfirmed = {
-                eventPublisher(LiveStreamContract.UiEvent.BookmarkStream(forceUpdate = true))
-            },
-            onClose = {
-                eventPublisher(LiveStreamContract.UiEvent.DismissBookmarkConfirmation)
-            },
-        )
-    }
+    ZapHost(
```
markocic, line 126: This is great!

app/src/main/kotlin/net/primal/android/stream/player/StreamState.kt
```diff
@@ -42,7 +42,14 @@ class StreamState internal constructor(
     }
 
     fun hide() {
-        _mode = StreamMode.Hidden(modeToRestore = _mode)
+        val naddr = when (_mode) {
+            is StreamMode.Minimized -> (_mode as StreamMode.Minimized).naddr
```
markocic: This is an unsafe cast. We should save `_mode` into a variable and rely on smart casting for this.

markocic (changes requested):
App is flying now!

app/src/main/kotlin/net/primal/android/stream/player/StreamState.kt
AleksandarIlic: 🫡

app/src/main/kotlin/net/primal/android/stream/player/StreamMode.kt
AleksandarIlic: Good point!

app/src/main/kotlin/net/primal/android/notes/feed/zaps/ZapHost.kt
AleksandarIlic: This is not a regression. It works as before. We can implement `rememberSaveable` when we migrate the entire codebase to use the `ZapHost`.

PR #604, opened 2025-08-20
Feature/stream pills

markocic:


AleksandarIlic (approved):
This is a great start. We might need to change the logic for the detection of live streams in feeds, but this could be done later.

AleksandarIlic (changes requested):
Actually, we need to fix a minor UI edge case before we merge this. New notes pill is overlapping the streaming row.

AleksandarIlic (approved):

PR #605, opened 2025-08-20
Feature/chat and zap details

mehmedalijaK:


AleksandarIlic (changes requested):
Can we extract bottom sheets or their contents from LiveStreamScreen into it's own file? LIveStreamScreen is getting very large.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -148,30 +157,83 @@ class LiveStreamViewModel @AssistedInject constructor(
...
+                }
+        }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
     private fun observeZaps() =
         viewModelScope.launch {
-            streamRepository.observeStream(aTag = streamNaddr.asATagValue())
-                .filterNotNull()
-                .map { it.eventZaps.map { zap -> StreamChatItem.ZapMessageItem(zap.asEventZapUiModel()) } }
+            activeAccountStore.activeUserAccount
```
AleksandarIlic: This might be unnecessary complication. Instead of embedding information `isZapperFollowed` and  `isZapperMuted` into chat list items, could we keep in the UiState the global list of muted and followed users (you already have for muted users) and then resolve this information on the UI level? This would simplify the viewModel logic and you can centralize observing active account on one place.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -258,10 +378,23 @@ class LiveStreamViewModel @AssistedInject constructor(
         }
     }
 
-    private fun String.replaceUserMentionsWithNostrUris(users: List<NoteTaggedUser>): String {
+    private suspend fun String.replaceUserMentionsWithNostrUris(users: List<NoteTaggedUser>): String {
```
AleksandarIlic: Can we avoid copy-pasting the code within our codebase!? This code is now identical to the one in `NoteEditorViewModel` and should be part of some helper/util class or something.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -148,30 +157,83 @@ class LiveStreamViewModel @AssistedInject constructor(
...
                 }
         }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
     private fun observeChatMessages() =
         viewModelScope.launch {
-            liveStreamChatRepository.observeMessages(streamATag = streamNaddr.asATagValue())
-                .map { chatList -> chatList.map { it.toChatMessageItem() } }
-                .collect {
+            activeAccountStore.activeUserAccount
```
AleksandarIlic: The same comment here (from zaps) applies here as well.

app/src/main/kotlin/net/primal/android/stream/ui/StreamDescriptionSection.kt
```diff
@@ -0,0 +1,172 @@
...
+                        val hashtag = it.item.substring(1)
+                        onHashtagClick(hashtag)
+                    }
+                },
+            )
+        }
+    }
+}
+
+private fun String.detectHashtags(): List<String> = Regex("""#\w+""").findAll(this).map { it.value }.toList()
```
AleksandarIlic: I think we already have a hashtag regex in HashtagUtils or something like that. Let's avoid duplicating the code.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -463,7 +448,19 @@ private fun LiveChatListOrSearch(
...
+    onZapMessageClick: (StreamChatItem.ZapMessageItem) -> Unit,
 ) {
+    if (state.chatLoading) {
+        Box(
+            modifier = modifier.fillMaxSize(),
+            contentAlignment = Alignment.Center,
+        ) {
+            PrimalLoadingSpinner()
+        }
+        return
```
AleksandarIlic: We should not do this in compose.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -463,7 +448,19 @@ private fun LiveChatListOrSearch(
     listState: LazyListState,
     eventPublisher: (LiveStreamContract.UiEvent) -> Unit,
     onProfileClick: (String) -> Unit,
+    onChatMessageClick: (StreamChatItem.ChatMessageItem) -> Unit,
+    onZapMessageClick: (StreamChatItem.ZapMessageItem) -> Unit,
 ) {
+    if (state.chatLoading) {
```
AleksandarIlic: I think I removed this because it was causing some UI issues. Can we remove it now? We are going to have additional task later to add proper loading indicators.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -969,3 +986,9 @@ private fun AnnotatedString.Builder.addProfileAnnotation(
         startIndex = content.indexOf(displayHandle, startIndex + 1)
     }
 }
+
+private data class ZapProfileHandlerArgs(
```
AleksandarIlic: Please remove this. You can always get `coroutineScope` and `context` in every composable function.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -213,55 +201,44 @@ fun LiveStreamScreen(
     )
 }
 
-@OptIn(ExperimentalMaterial3Api::class)
 @Composable
-private fun LiveStreamModalBottomSheets(
-    activeSheet: ActiveBottomSheet,
-    onDismiss: () -> Unit,
+private fun LiveStreamScreenSideEffects(
```
AleksandarIlic: Please remove this function and return the composable in the `LiveStreamScreen`. `ZapHost` and `FollowsApprovalAlertDialog` are not side effects. If you need to encapsulate something, that could be the `Scaffold` composable which could be `private fun LiveScreenScaffold()`

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamBottomSheet.kt
```diff
@@ -0,0 +1,379 @@
...
+        derivedStateOf {
+            val streamInfo = state.streamInfo
+            val activeUserId = state.activeUserId
+            if (streamInfo == null || activeUserId == null) {
+                return@derivedStateOf null
+            }
+
+            when (activeSheet) {
+                ActiveBottomSheet.StreamInfo -> BottomSheetProfile(
+                    details = streamInfo.mainHostProfile!!,
```
markocic: We should avoid `!!` as that crashes the app.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamBottomSheet.kt
```diff
@@ -0,0 +1,379 @@
...
+data class BottomSheetProfile(
+    val details: ProfileDetailsUi,
+    val stats: ProfileStatsUi?,
+    val isFollowed: Boolean,
+    val isMuted: Boolean,
+    val isActiveUser: Boolean,
+)
+
+@Composable
+private fun rememberBottomSheetProfile(
```
markocic: This function seems off to me. Why do we need to perform so much calculations on UI? I don't understand why we are filtering messages/zaps and then using the one from `activeSheet` as fallback?

app/src/main/kotlin/net/primal/android/stream/ui/StreamChatItem.kt
```diff
@@ -8,13 +10,20 @@ sealed interface StreamChatItem {
...
+        val isAuthorMuted: Boolean,
+        val authorProfileStats: ProfileStatsUi?,
     ) : StreamChatItem {
         override val timestamp: Long = message.timestamp
         override val uniqueId: String = message.messageId
     }
 
     data class ZapMessageItem(
         val zap: EventZapUiModel,
+        val zapperProfile: ProfileDetailsUi?,
```
markocic: This seems unnecessary. Isn't profile information in `EventZapUiModel` enough?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -700,17 +821,24 @@ class LiveStreamViewModel @AssistedInject constructor(
                     Timber.w(error, "Failed to resolve profile for $uriString")
                     null
                 }
+            } else {
+                null
             }
         }
 
+        val stats = profileRepository.findProfileStats(profileIds = listOf(this.author.profileId)).firstOrNull()
```
markocic: This could be resolved when getting message information. Adding `ProfileStats` connection in `StreamChatMessage` would do the trick.

markocic (changes requested):

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -120,6 +124,8 @@ class LiveStreamViewModel @AssistedInject constructor(
 
     private var zaps: List<StreamChatItem.ZapMessageItem> = emptyList()
     private var chatMessages: List<StreamChatItem.ChatMessageItem> = emptyList()
+    private var initialZapsReceived = false
+    private var initialMessagesReceived = false
```
AleksandarIlic: Why do we need this flags? Can we remove them?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -272,10 +338,23 @@ class LiveStreamViewModel @AssistedInject constructor(
         }
     }
 
-    private fun String.replaceUserMentionsWithNostrUris(users: List<NoteTaggedUser>): String {
+    private suspend fun String.replaceUserMentionsWithNostrUris(users: List<NoteTaggedUser>): String {
```
AleksandarIlic: This is still duplicated code from NoteEditorViewModel as well. Can we have this logic in 1 place?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -229,10 +265,40 @@ class LiveStreamViewModel @AssistedInject constructor(
...
+                val statsList = profileRepository.findProfileStats(profileIds = listOf(profileId))
+                val followerCount = statsList.firstOrNull()?.followers
+
+                if (followerCount != null) {
+                    setState {
+                        val newMap = this.profileIdToFollowerCount.toMutableMap()
+                        newMap[profileId] = followerCount
+                        copy(profileIdToFollowerCount = newMap)
+                    }
+                }
```
markocic: Converting to mutable map should be avoided here. Optimally we should do:
```kt
setState {
    copy(profileIdToFollowerCount = profileIdToFollowerCount + (profileId to followerCount))
}
```

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -339,8 +418,8 @@ class LiveStreamViewModel @AssistedInject constructor(
         authorObserversJob = viewModelScope.launch {
             observeAuthorProfile(mainHostId)
             observeAuthorProfileStats(mainHostId)
-            observeFollowState(mainHostId)
-            observeMuteState(mainHostId)
+            observeFollowState()
+            observeMuteState()
```
markocic: Why are these dependent on `mainHost`? Could we just observe this in the `init`?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -372,30 +451,30 @@ class LiveStreamViewModel @AssistedInject constructor(
                 }
         }
 
-    private fun CoroutineScope.observeFollowState(mainHostId: String) =
+    private fun CoroutineScope.observeFollowState() =
         launch {
             activeAccountStore.activeUserAccount
-                .map { mainHostId in it.following }
                 .distinctUntilChanged()
```
markocic: We already have `distinctUntilChanged` on `activeUserAccount`.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -372,30 +451,30 @@ class LiveStreamViewModel @AssistedInject constructor(
...
-            mutedItemRepository.observeIsUserMutedByOwnerId(
-                pubkey = mainHostId,
-                ownerId = activeAccountStore.activeUserId(),
-            ).collect {
-                setState {
-                    copy(
-                        streamInfo = this.streamInfo?.copy(isMainHostMutedByActiveUser = it),
-                    )
+            mutedItemRepository.observeMutedUsersByOwnerId(ownerId = activeAccountStore.activeUserId())
+                .distinctUntilChanged()
```
markocic: We already have `distinctUntilChanged` on `activeUserAccount`.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -523,13 +599,6 @@ class LiveStreamViewModel @AssistedInject constructor(
             profileFollowsHandler.observeResults().collect {
                 when (it) {
                     is ProfileFollowsHandler.ActionResult.Error -> {
```
markocic, line 588: On `Error` we should now reverse all actions. Take a look at `FollowPackViewModel:176` to see how that is done.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -646,6 +721,31 @@ class LiveStreamViewModel @AssistedInject constructor(
...
+                profileId = authorId,
+                eventId = messageId,
+                articleId = streamInfo.atag,
+            )
+        } catch (error: SigningKeyNotFoundException) {
+            Timber.w(error)
+            setState { copy(error = UiError.MissingPrivateKey) }
+        } catch (error: SigningRejectedException) {
+            Timber.w(error)
+            setState { copy(error = UiError.NostrSignUnauthorized) }
```
markocic: This should be compacted to catch `SignatureException` and use a mapper to resolve `UiError` from it.

I can see that this was done at couple other places as well. I guess we missed it. Could you update them as well?

app/src/main/kotlin/net/primal/android/stream/ui/ChatDetailsSection.kt
```diff
@@ -0,0 +1,106 @@
...
+        ) {
+            Row(
+                verticalAlignment = Alignment.CenterVertically,
+            ) {
+                Icon(
+                    modifier = Modifier.size(20.dp),
+                    imageVector = PrimalIcons.ContextReportContent,
+                    contentDescription = null,
+                )
+                Spacer(modifier = Modifier.width(8.dp))
```
markocic: There is no need for `Spacer` inside `Row`. We can use `Arrangement.spacedBy(8.dp)` to get the same effect. This applies to all occurrences of this.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamBottomSheets.kt
```diff
@@ -0,0 +1,228 @@
...
+    onHashtagClick: (String) -> Unit,
+    onMessageClick: (String) -> Unit,
+    onEditProfileClick: () -> Unit,
+    onDrawerQrCodeClick: (String) -> Unit,
+) {
+    val sheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)
+
+    val profileIdToFetch = activeSheet.getProfileId(streamInfo)
+    LaunchedEffect(profileIdToFetch) {
+        if (profileIdToFetch != null && !followerCountMap.containsKey(profileIdToFetch)) {
```
markocic: I believe we are already checking in ViewModel if map contains this key. We should stick to one or the other here.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamBottomSheets.kt
```diff
@@ -0,0 +1,228 @@
...
+import net.primal.android.core.compose.profile.model.ProfileDetailsUi
+import net.primal.android.events.ui.EventZapUiModel
+import net.primal.android.profile.details.ui.model.PremiumProfileDataUi
+import net.primal.android.stream.LiveStreamContract
+import net.primal.android.theme.AppTheme
+import net.primal.domain.nostr.ReportType
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun LiveStreamModalBottomSheetHost(
```
markocic, line 34: Design colors are different then the one here. Notice the bottom section when bottom bar opens. Not sure if the design changed in the meantime but this should be adjusted.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -139,10 +134,17 @@ fun LiveStreamScreen(
 ) {
     val context = LocalContext.current
     val snackbarHostState = remember { SnackbarHostState() }
+    var activeBottomSheet by remember { mutableStateOf<ActiveBottomSheet>(ActiveBottomSheet.None) }
```
markocic: I think this should be part of `UiState`. We are performing action every time this changes. Right now UI decides when follower count should be fetched. UI should be concerned with stuff like that. By moving this to UiState we are going thru `ViewModel` to change it and we can perform side effects like fetching followers count.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -760,22 +793,29 @@ private fun ChatMessageListItem(message: ChatMessageUi, onProfileClick: (String)
...
                     return@PrimalClickableText
                 }
+
+                onClick()
             },
         )
     }
 }
 
+@OptIn(ExperimentalFoundationApi::class)
```
markocic: Is this necessary?

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -688,8 +714,14 @@ private fun LiveChatCommentInput(
...
```
markocic: Is this necessary?

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -901,8 +941,7 @@ private fun handleZapProfile(
             snackbarHostState.showSnackbar(
                 message = context.getString(
                     R.string.wallet_send_payment_error_nostr_user_without_lightning_address,
-                    profileDetails?.authorDisplayName
-                        ?: context.getString(R.string.wallet_send_payment_this_user_chunk),
+                    profileDetails.authorDisplayName,
                 ),
                 duration = SnackbarDuration.Short,
             )
```
markocic, line 957: This message isn't shown because it is covered by `BottomSheet`.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -879,20 +919,20 @@ private fun ZapMessageContent(zap: EventZapUiModel) {
 }
 
 private fun handleZapProfile(
-    state: LiveStreamContract.UiState,
+    profileDetails: ProfileDetailsUi?,
```
markocic: Make this non-nullable as we are returning immediately if it is `null`.

app/src/main/kotlin/net/primal/android/stream/ui/StreamDescriptionSection.kt
markocic: Is there a reason we are not using `parseHashtags` from `HashtagUtils`?

markocic (changes requested):

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -171,6 +172,9 @@ class LiveStreamViewModel @AssistedInject constructor(
                 .filterNotNull()
                 .map { it.eventZaps.map { zap -> StreamChatItem.ZapMessageItem(zap.asEventZapUiModel()) } }
                 .collect {
+                    if (!initialZapsReceived) {
+                        initialZapsReceived = true
+                    }
```
markocic: We should just do `initialZapsReceived = true`. This applies to messages as well. What happens when we don't any zaps but we do have messages? Will they be loaded?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -247,10 +278,50 @@ class LiveStreamViewModel @AssistedInject constructor(
                     }
 
                     is UiEvent.ChangeContentModeration -> changeContentModeration(moderationMode = it.moderationMode)
+
+                    is UiEvent.ReportMessage -> reportMessage(
+                        reportType = it.reportType,
+                        messageId = it.messageId,
+                        authorId = it.authorId,
+                    )
+                    is UiEvent.FetchFollowerCount -> fetchFollowerCount(profileId = it.profileId)
```
markocic: Do we need this event now? If not, we should delete it along with the `LaunchedEffect` in UI that calls this event.

markocic (changes requested):
Did a quick check. Some of the previous comments weren't resolved. I marked the rest as resolved.

markocic (approved):

PR #620, opened 2025-08-27
Implement stream player loading indicators

markocic:


app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayer.kt
```diff
@@ -58,8 +57,9 @@ fun LiveStreamPlayer(
     onToggleFullScreenClick: () -> Unit,
     modifier: Modifier = Modifier,
     playerModifier: Modifier = Modifier,
+    loadingModifier: Modifier = Modifier,
```
AleksandarIlic, line 60: Do we really need two modifiers?

AleksandarIlic (approved):

PR #622, opened 2025-08-27
Feature/recording streams

mehmedalijaK:


app/src/main/res/values/strings.xml
```diff
@@ -1368,6 +1368,7 @@
     <string name="live_stream_chat_message">Chat Message</string>
     <string name="live_stream_bottom_sheet_unmute_user">Unmute</string>
     <string name="live_stream_bottom_sheet_mute_user">Mute</string>
+    <string name="live_stream_recording_not_available">Video recording not available</string>
```
AleksandarIlic: Let's go actually with: "Video not available"

app/src/main/kotlin/net/primal/android/stream/LiveStreamContract.kt
```diff
@@ -104,6 +105,7 @@ interface LiveStreamContract {
             val messageId: String,
             val authorId: String,
         ) : UiEvent()
+        data object OnPlayerUnavailable : UiEvent()
```
AleksandarIlic: Player is there, we just can play the video. Maybe better name would be `OnVideoUnavailable` or something like that.

AleksandarIlic (dismissed):

AleksandarIlic (approved):

PR #623, opened 2025-08-28
Feature/stream settings bottom sheet

markocic:


app/src/main/kotlin/net/primal/android/stream/LiveStreamContract.kt
```diff
@@ -36,6 +36,7 @@ interface LiveStreamContract {
         val userTaggingState: UserTaggingState = UserTaggingState(),
         val error: UiError? = null,
         val chatLoading: Boolean = true,
+        val isStreamMuted: Boolean = false,
```
AleksandarIlic: I think it should be `mainHostStreamsMuted` since this mute affects all host's streams. And this should apply for event as well I guess. Or am I over complicating it?

data/caching/local/src/commonMain/kotlin/net/primal/data/local/dao/mutes/MutedItemData.kt
```diff
@@ -9,4 +9,5 @@ data class MutedItemData(
     val item: String,
     val ownerId: String,
     val type: MutedItemType,
+    val listType: ListType,
```
AleksandarIlic, line 12: Why is this necessary?

AleksandarIlic (approved):

PR #625, opened 2025-08-29
Implement live notifications

mehmedalijaK:


data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/mappers/remote/ContentPrimalNotification.kt
```diff
@@ -26,4 +26,7 @@ data class ContentPrimalNotification(
     @SerialName("who_highlighted_it") val whoHighlightedIt: String? = null,
     @SerialName("who_bookmarked_it") val whoBookmarkedIt: String? = null,
     val reaction: String? = null,
+    val host: String? = null,
+    @SerialName("live_event_id") val liveEventId: String? = null,
+    val coordinate: String? = null,
```
AleksandarIlic: Is this `aTag` actually?

domain/primal/src/commonMain/kotlin/net/primal/domain/notifications/NotificationSettingsType.kt
```diff
@@ -141,5 +141,13 @@ sealed class NotificationSettingsType(val id: String, val order: Int) {
             ),
             order = 6,
         )
+
+        data object LiveEvents : TabNotifications(
+            id = "LIVE_EVENTS",
```
AleksandarIlic: Actually, we should not implement this, it could cause issues with other platforms. Forgot to mention that, sorry. Could you comment this out so we can easily enable it later?

AleksandarIlic (approved):

PR #630, opened 2025-09-02
Pause player when app enters background

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/LiveStreamOverlay.kt
```diff
@@ -117,6 +121,20 @@ private fun LiveStreamOverlay(
         },
     )
 
+    val lifecycleOwner = LocalLifecycleOwner.current
+    DisposableEffect(lifecycleOwner, exoPlayer) {
```
AleksandarIlic: I think we should use official lifecycle Effects from Compose:

        LifecycleStartEffect(keys) {
            onStopOrDispose {
                // Handle ON_STOP event or DISPOSE
            }
        }

AleksandarIlic (changes requested):

PR #632, opened 2025-09-02
Feature/wallet tx metadata

markocic:


AleksandarIlic (dismissed):

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/events/EventRepositoryImpl.kt
```diff
@@ -186,6 +192,35 @@ class EventRepositoryImpl(
...
+                if (invoice != null && nostrEvent != null) invoice to nostrEvent else null
+            }.toMap()
+
+            val missingZapReceiptsByInvoice = invoices.toSet() - localZapReceipts.mapNotNull { it.invoice }.toSet()
+
+            if (missingZapReceiptsByInvoice.isNotEmpty()) {
+                eventStatsApi.getZapReceipts(invoices = missingZapReceiptsByInvoice.toList())
+                    .mapCatching { response ->
+                        val map = response.mapEvent?.content?.decodeFromJsonStringOrNull<Map<String, NostrEvent>>()
+                            ?: throw IllegalArgumentException("failed to parse invoices map.")
```
AleksandarIlic, line 212: This is not crashing the app, right? Why not using `mapNotNull` and return `null` if we can't decode it? This is how is done across the app I think.

AleksandarIlic (approved):

PR #639, opened 2025-09-04
Improve `FullScreenController`

markocic:
We now remember the last `requestedOrientation` and restoring it once we are done with `FullScreenController`. This allows other screens to keep their preferred orientation.

`isPortraitDeg` got a sagnificant improvement once we discovered that mapping of rotation degrees to orientation depends on the current orientation. Android System favors the current orientation in this resolution process.

We are now keeping track of auto rotate system setting. Small but key change was adding `delay` before unlocking rotation to allow the device to settle. Some weird rotation bugs occur if user rotates the device slowly.

### DIFF
--- a/app/src/main/kotlin/net/primal/android/core/compose/FullScreenController.kt
+++ b/app/src/main/kotlin/net/primal/android/core/compose/FullScreenController.kt
@@ -2,8 +2,10 @@ package net.primal.android.core.compose
 
 import android.annotation.SuppressLint
 import android.app.Activity
+import android.content.Context
 import android.content.pm.ActivityInfo
 import android.content.res.Configuration
+import android.provider.Settings
 import android.view.OrientationEventListener
 import androidx.activity.compose.BackHandler
 import androidx.annotation.MainThread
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/net/primal/android/core/compose/FullScreenController.kt
```diff
@@ -2,50 +2,78 @@ package net.primal.android.core.compose
...
-                unlockRotation()
-            }
-
-            if (isFullscreen != !isPortraitDeg(orientation)) {
-                isFullscreen = !isPortraitDeg(orientation)
-                onFullscreenChanged(isFullscreen)
+            if (context.isAutoRotateEnabled()) {
+                if (!isFullscreen && !hasUnlockedAfterPortrait && isPortraitDeg(orientation)) {
+                    scope.launch {
+                        delay(1.seconds)
```
AleksandarIlic, line 66: We could have this time as a const `UNLOCK_DELAY` or something since it's used in two places.

PR #641, opened 2025-09-04
Implement chat and zap loading logic

markocic:


app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -185,13 +191,16 @@ class LiveStreamViewModel @AssistedInject constructor(
         }
 
     private fun updateChatItems() {
+        val zaps = zaps ?: return
+        val chatMessages = chatMessages ?: return
```
AleksandarIlic: This looks strange.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -147,7 +149,7 @@ class LiveStreamViewModel @AssistedInject constructor(
     private fun changeContentModeration(moderationMode: StreamContentModerationMode) =
         viewModelScope.launch {
             if (moderationMode != _state.value.contentModerationMode) {
-                setState { copy(loading = true, contentModerationMode = moderationMode) }
+                setState { copy(chatLoading = true, contentModerationMode = moderationMode) }
```
AleksandarIlic: We should probably remove `chatLoading` and use only `loading`. Also we should change `streamRepository::startLiveStreamSubscription` batching logic to process events after some amount of inactivity and not on every second. This should remove the jumps on stream cold start.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -165,11 +167,15 @@ class LiveStreamViewModel @AssistedInject constructor(
 
     private fun observeZaps() =
         viewModelScope.launch {
-            streamRepository.observeStream(aTag = streamNaddr.asATagValue())
-                .filterNotNull()
-                .map { it.eventZaps.map { zap -> StreamChatItem.ZapMessageItem(zap.asEventZapUiModel()) } }
-                .collect {
-                    zaps = it
+            eventRepository.observeZapsByEventId(eventId = streamNaddr.asATagValue())
```
AleksandarIlic, line 171: It says `eventId`, but we are sending `aTag`. And it works. We should probably fix the namings here since we are in the neighbourhood.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -66,18 +70,30 @@ class StreamRepositoryImpl(
...
-        liveStreamApi.subscribeToLiveEvent(
-            streamingNaddr = naddr,
-            userId = userId,
-            contentModerationMode = when (streamContentModerationMode) {
-                StreamContentModerationMode.Moderated -> "moderated"
-                StreamContentModerationMode.None -> "all"
-            },
-        ).collect { response ->
-            processLiveStreamResponse(response = response)
+    ): Job =
```
AleksandarIlic, line 73: Let's go with `awaitLiveStreamSubscriptionStart`.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -135,8 +137,8 @@ class LiveStreamViewModel @AssistedInject constructor(
 
     private fun startLiveStreamSubscription() {
         streamSubscriptionJob?.cancel()
-        streamSubscriptionJob = viewModelScope.launch {
-            streamRepository.startLiveStreamSubscription(
+        viewModelScope.launch {
```
AleksandarIlic, line 139: Let's put cancel as well in the launch block.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -147,10 +149,14 @@ class LiveStreamViewModel @AssistedInject constructor(
     private fun changeContentModeration(moderationMode: StreamContentModerationMode) =
         viewModelScope.launch {
             if (moderationMode != _state.value.contentModerationMode) {
-                setState { copy(loading = true, contentModerationMode = moderationMode) }
                 val aTagValue = streamNaddr.asATagValue()
+
+                streamSubscriptionJob?.cancel()
+                streamSubscriptionJob = null
```
AleksandarIlic: Is this really necessary? We are calling `startLiveStreamSubscription` which is going to cancel it and replace it with new job.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -187,11 +197,13 @@ class LiveStreamViewModel @AssistedInject constructor(
     private fun updateChatItems() {
         val mutedProfiles = state.value.activeUserMutedProfiles
 
-        val filteredZaps = zaps.filterNot { it.zap.zapperId in mutedProfiles }
-        val filteredChatMessages = chatMessages.filterNot { it.message.authorProfile.pubkey in mutedProfiles }
+        val filteredZaps = zaps?.filterNot { it.zap.zapperId in mutedProfiles } ?: return
+        val filteredChatMessages = chatMessages
+            ?.filterNot { it.message.authorProfile.pubkey in mutedProfiles }
+            ?: return
```
AleksandarIlic, line 200: You have two returns here. What would happen if there are no zaps, but there are chat messages? Or opposite.

data/caching/remote/src/commonMain/kotlin/net/primal/data/remote/api/stream/LiveStreamApiImpl.kt
```diff
@@ -44,8 +45,8 @@ class LiveStreamApiImpl(
                         contentModerationMode = contentModerationMode,
                     ).encodeToJsonString(),
                 ),
+                inactivityTimeout = 500.milliseconds,
```
AleksandarIlic, line 48: Are we safe without catch?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
markocic: That just means zaps or messages didn't emit anything from the flow. That is the desired result.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
markocic: Yes. Cancelling and then setting to `null` will make sure to render `chatLoading` state on the UI.

data/caching/remote/src/commonMain/kotlin/net/primal/data/remote/api/stream/LiveStreamApiImpl.kt
markocic: We have `catch` inside `StreamRepository` from which it's called so yes.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
markocic: Now that I think about, I could use `isActive` to detect this without the need of explicitly setting `null`. Thanks for the idea!

PR #650, opened 2025-09-09
Improve video unavailable handling

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamMiniPlayer.kt
```diff
@@ -245,10 +251,17 @@ private fun PlayerControls(
...
-    Box(
-        modifier = Modifier
+
+    val boxModifier = if (isStreamUnavailable) {
+        Modifier.fillMaxSize()
+    } else {
+        Modifier
             .fillMaxSize()
-            .clickable { onExpandStream() },
+            .clickable { onExpandStream() }
```
markocic: We should disable `clickable` instead of changing `Modifier` applied. If we change `Modifier` like this, we are triggering unnecessary recompositions.

app/src/main/kotlin/net/primal/android/stream/LiveStreamContract.kt
```diff
@@ -56,7 +56,7 @@ interface LiveStreamContract {
         val isLive: Boolean = false,
         val isVideoFinished: Boolean = false,
     ) {
-        val isLoading: Boolean get() = isBuffering && !isPlaying
+        val isLoading: Boolean get() = isBuffering
```
markocic, line 59: Why was this change necessary? If it was required, could we just use `isBuffering` instead of having alias variable?

markocic (changes requested):
I didn't test, just skimmed through the PR. Wanted to check for usages of `mediaController`. All looks good in that regard!

app/src/main/kotlin/net/primal/android/stream/LiveStreamContract.kt
AleksandarIlic: I see this comment not resolved? Why is this change necessary?

PR #651, opened 2025-09-09
Fix generating share stream link

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/utils/ShareUtils.kt
```diff
@@ -70,9 +69,26 @@ fun resolvePrimalArticleLink(
...
+        "https://primal.net/$nprofile/live/$identifier"
+    }
+}
 
 fun resolvePrimalProfileLink(profileId: String, primalName: String?): String {
     val path = primalName ?: "p/${Nprofile(pubkey = profileId).toNprofileString()}"
     return "https://primal.net/$path"
 }
+
+private fun extractPrimalName(internetIdentifier: String?): String? {
```
AleksandarIlic: The original implementation was not correct. PrimalName should come from the database (where is verified) and not from nip05 (where is unverified).

AleksandarIlic (approved):

PR #656, opened 2025-09-10
Implement `UpdateStaleStreamDataUseCase` and invoke it in `HomeFeedViewModel`

markocic:


app/src/main/kotlin/net/primal/android/core/di/UseCasesModule.kt
```diff
@@ -0,0 +1,23 @@
...
+import net.primal.domain.usecase.UpdateStaleStreamDataUseCase
+
+@Module
+@InstallIn(SingletonComponent::class)
+object UseCasesModule {
+    @Provides
+    fun provideUpdateStaleStreamDataUseCase(
+        streamRepository: StreamRepository,
+        eventRepository: EventRepository,
+    ): UpdateStaleStreamDataUseCase =
```
AleksandarIlic: This should probably belong to some Stream DI module. Do we have it?

app/src/main/kotlin/net/primal/android/notes/home/HomeFeedViewModel.kt
```diff
@@ -139,6 +142,8 @@ class HomeFeedViewModel @Inject constructor(
         }
     }
 
+    private fun updateStaleStreamData() = viewModelScope.launch { updateStaleStreamDataUseCase.invoke() }
```
AleksandarIlic: I'm not sure if we should call this on `init` or do some debounce logic like we have for user data updates.

AleksandarIlic (changes requested):
Let's take the time to create `DataUpdater` which should take away and handle all data updates from `HomeFeedViewModel`. This would simplify the viewmodel's depenedencies.

app/src/main/kotlin/net/primal/android/core/di/UseCasesModule.kt
markocic: No. Stream DI is handled in `CachingRepositoriesModule`. Not sure if we should separate it.

AleksandarIlic: Ok, so there is no stream DI module in `app` module. I don't like the generic `UseCasesModule`. I would prefer per domain modules, maybe you should create a Stream DI module in `app` module?

app/src/main/kotlin/net/primal/android/navigation/splash/SplashViewModel.kt
```diff
@@ -34,6 +34,6 @@ class SplashViewModel @Inject constructor(
 
     private fun fetchLatestAppConfig() =
         viewModelScope.launch {
-            appConfigHandler.updateAppConfigOrFailSilently()
+            appConfigHandler.invokeImmediately()
```
AleksandarIlic: It's not clear `appConfigHandler.invokeImmediately()` is doing. What are we invoking here? I'm I know what it is, but someone who looks for the first time will have to go into the function to understand it. We should either come up with better names or to wrap them in handler's specific function. Or do you have a better idea?

`invoke` functions makes sense for use cases which have a single action. Handlers have multiple. That's why it's confusing.

core/app-config/src/commonMain/kotlin/net/primal/core/config/AppConfigHandler.kt
```diff
@@ -1,14 +1,60 @@
...
+import kotlinx.coroutines.withContext
+import net.primal.core.config.api.ApiConfigResponse
+import net.primal.core.config.api.WellKnownApi
+import net.primal.core.config.store.AppConfigDataStore
+import net.primal.core.utils.Result
+import net.primal.core.utils.coroutines.DispatcherProvider
+import net.primal.core.utils.debouncer.Debouncer
+import net.primal.domain.common.exception.NetworkException
 
-interface AppConfigHandler {
```
AleksandarIlic, line 5: Why is there an `interface` for this? Does the git history says anything useful?

core/utils/src/commonMain/kotlin/net/primal/core/utils/debouncer/Debouncer.kt
```diff
@@ -0,0 +1,33 @@
+package net.primal.core.utils.debouncer
+
+import kotlin.time.Duration
+import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
+import net.primal.core.utils.Result
+
+abstract class Debouncer {
```
AleksandarIlic: Maybe `Updater`?
`canUpdate`?
`updateImmediately`?
`updateWithDebounce`?
`doUpdate`?

This might solve the issue with handlers. Could also go into the `updater` package. It does not have to generic debouncer.

app/src/main/kotlin/net/primal/android/core/updater/DataUpdater.kt
```diff
@@ -0,0 +1,48 @@
...
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
+import net.primal.android.user.accounts.active.ActiveAccountStore
+import net.primal.android.user.updater.UserDataUpdater
+import net.primal.android.user.updater.UserDataUpdaterFactory
+import net.primal.core.config.AppConfigHandler
+import net.primal.core.utils.coroutines.DispatcherProvider
+import net.primal.domain.usecase.UpdateStaleStreamDataUseCase
+
+class DataUpdater @Inject constructor(
```
AleksandarIlic, line 17: Is this created as Singleton or not? If not, then every instance will have unbounded `scope` which means `observeActiveAccount` coroutines will never finish.

PR #658, opened 2025-09-11
Setup cronet in PrimalMediaSessionService

mehmedalijaK:


app/build.gradle.kts
```diff
@@ -342,6 +342,8 @@ dependencies {
     implementation(libs.media3.exoplayer.hls)
     implementation(libs.media3.exoplayer.ui)
     implementation(libs.media3.exoplayer.ui.compose)
+    implementation(libs.media3.datasource.cronet)
+    implementation(libs.google.play.services.cronet)
```
AleksandarIlic: I think we are going to have to include this for Google flavour only. This probably has some analytics and tracking under the hood.

app/src/main/kotlin/net/primal/android/core/service/PrimalMediaSessionService.kt
```diff
@@ -8,34 +8,179 @@ import androidx.core.net.toUri
...
             .setAudioAttributes(audioAttributes, true)
-            .build()
 
-        mediaSession = MediaSession.Builder(this, player)
+        cronetEngine = runCatching {
+            CronetEngine.Builder(this)
+                .enableHttp2(true)
+                .enableQuic(true)
+                .enableBrotli(true)
+                .setUserAgent("PrimalAndroid/${BuildConfig.VERSION_NAME}")
```
markocic: Might be good to use `UserAgentProvider.APP_NAME` instead of hard-coding `PrimalAndroid`.

app/src/main/kotlin/net/primal/android/core/service/PrimalMediaSessionService.kt
```diff
@@ -8,34 +8,179 @@ import androidx.core.net.toUri
...
+            }
+            analyticsListener.setCronetInfo(version)
+
+            Timber.i("Cronet enabled for media playback: $version")
+        }.onFailure {
+            Timber.w(it, "Cronet is not available. Using default HTTP stack for media playback.")
+        }.getOrNull()
+
+        val player = playerBuilder.build()
+        player.addAnalyticsListener(analyticsListener)
```
markocic: What happens if `cronetEngine` fails to be initialized?

app/src/main/kotlin/net/primal/android/core/service/PrimalMediaSessionService.kt
```diff
@@ -8,34 +8,179 @@ import androidx.core.net.toUri
...
+                .enableBrotli(true)
+                .setUserAgent("PrimalAndroid/${BuildConfig.VERSION_NAME}")
+                .build()
+        }.onSuccess { engine ->
+            val executor = Executors.newSingleThreadExecutor()
+            this.cronetExecutor = executor
+
+            val dataSourceFactory = CronetDataSource.Factory(engine, executor)
+            val mediaSourceFactory = DefaultMediaSourceFactory(this).setDataSourceFactory(dataSourceFactory)
+            playerBuilder.setMediaSourceFactory(mediaSourceFactory)
```
markocic: Do we need some `MediaSourceFactory` in the `onFailure` as well? Does it have some default under the hood we are falling back to?

app/src/main/kotlin/net/primal/android/core/service/PrimalMediaSessionService.kt
mehmedalijaK: If the cronetEngine fails to be initialized the application will fall back to using ExoPlayer's default HTTP stack for media playback.

app/src/main/kotlin/net/primal/android/core/service/PrimalMediaSessionService.kt
mehmedalijaK: No, you do not need to set a MediaSourceFactory in the onFailure block because `ExoPlayer.Builder` will create a default one for you under the hood.

app/src/google/kotlin/net/primal/android/core/service/PrimalMediaSessionService.kt
```diff
@@ -0,0 +1,263 @@
...
+import java.io.IOException
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import net.primal.android.MainActivity
+import net.primal.android.networking.UserAgentProvider
+import net.primal.core.utils.AndroidBuildConfig
+import org.chromium.net.CronetEngine
+import timber.log.Timber
+
+class PrimalMediaSessionService : MediaSessionService() {
```
markocic: I am not sure about this approach. We now have a lot of duplicate code and whoever changes this later should remember to copy-paste the code to google flavor as well. I think we should extract only the cronet logic into flavor specific code and provide it with DI. Similar think has been done so we can follow that approach here.

app/src/google/kotlin/net/primal/android/core/player/GooglePlayerManager.kt
```diff
@@ -0,0 +1,77 @@
...
+            this.cronetExecutor = executor
+
+            val dataSourceFactory = CronetDataSource.Factory(engine, executor)
+            val mediaSourceFactory = DefaultMediaSourceFactory(context).setDataSourceFactory(dataSourceFactory)
+            playerBuilder.setMediaSourceFactory(mediaSourceFactory)
+
+            val version = try {
+                engine.versionString
+            } catch (_: Exception) {
+                "unknown"
```
AleksandarIlic: Let's use `runCatching {}.getOrElse { "unknown" } `. `runCatching` has better API.

app/src/google/kotlin/net/primal/android/core/player/GooglePlayerManager.kt
```diff
@@ -0,0 +1,77 @@
...
+            }
+            analyticsListener.setCronetInfo(version)
+
+            Timber.i("Cronet enabled for media playback: $version")
```
AleksandarIlic: This should be removed.

app/src/main/kotlin/net/primal/android/core/service/PrimalMediaSessionService.kt
```diff
@@ -5,37 +5,40 @@ import android.content.Context
...
     private var mediaSession: MediaSession? = null
 
     override fun onCreate() {
-        super.onCreate()
-
-        val audioAttributes = AudioAttributes.Builder()
-            .setUsage(C.USAGE_MEDIA)
-            .setContentType(C.AUDIO_CONTENT_TYPE_MOVIE)
-            .build()
+        val entryPoint = EntryPointAccessors.fromApplication(
```
AleksandarIlic: You can annotate `PrimalMediaSessionService` with `@AndroidEntryPoint` and do field injection, just like in `Activity`.

PR #661, opened 2025-09-12
Resolve video rendering glitch during screen transitions

mehmedalijaK:


markocic (commented):
This looks okay to me. Not sure if there is a simpler way to achieve same result. If this is the solution, it might be good to wrap it in a reusable component so we don't introduce this bug later.

AleksandarIlic (approved):

PR #662, opened 2025-09-15
Add fallback item keys for media items in `PremiumMediaManagementScreen`

markocic:


app/src/main/kotlin/net/primal/android/premium/manage/media/PremiumMediaManagementScreen.kt
```diff
@@ -140,7 +140,7 @@ fun PremiumMediaManagementScreen(
 
             items(
                 items = state.mediaItems,
-                key = { it.mediaId },
+                key = { it.mediaId.ifBlank { it.mediaUrl }.ifBlank { it.hashCode() } },
```
AleksandarIlic: This is not going to work because `mediaId` is the same as `mediaUrl`. Looks like there is no stable id. I'm thinking of removing the key from this lazy column since we can not guarantee it. Could you give it a test without key? If there are no other issues, we can go with that approach.

PR #663, opened 2025-09-16
Implement seeking for recorded streams and adjust buffer

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayerControls.kt
```diff
@@ -346,111 +387,160 @@ private fun PlayerActionButtons(
     }
 }
 
-@OptIn(ExperimentalMaterial3Api::class)
 @Composable
-private fun PlayerSlider(
+private fun PrimalSeekBar(
```
AleksandarIlic: Can we extract this component out of this file? If it's generic, maybe it can be in `core.compose`.

AleksandarIlic (changes requested):
UX is much better now, but there is still room for improvement.

AleksandarIlic (approved):

PR #664, opened 2025-09-17
Fix chat scrolling automatically when new message or zap

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -644,8 +645,14 @@ private fun LiveChatContent(
 ) {
     val coroutineScope = rememberCoroutineScope()
 
-    LaunchedEffect(state.chatItems.size) {
-        if (state.chatItems.isNotEmpty()) {
+    val isUserAtBottom by remember {
```
AleksandarIlic: `isUserAtBottom` might not be the best name. If we flip the orientation, it's not going to make sense. Maybe the name could be `isChatShowingLatestContent` or something like that.

PR #665, opened 2025-09-17
Keep screen on during video playback

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayer.kt
```diff
@@ -81,6 +82,10 @@ fun LiveStreamPlayer(
     var controlsVisible by remember { mutableStateOf(false) }
     var menuVisible by remember { mutableStateOf(false) }
 
+    if (state.playerState.isPlaying) {
```
AleksandarIlic: This is not going to work. If you pause the video, the screen will still stay on, because inside `KeepScreenOn` the disposable effect takes `Unit` as key. You could create additional `KeepScreenOn` which could take `state.playerState.isPlaying` as key and use it in underlying `DisposableEffect`.

AleksandarIlic (approved):

PR #666, opened 2025-09-22
Implement optimistic updates for zaps

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -614,6 +623,28 @@ class LiveStreamViewModel @AssistedInject constructor(
         }
     }
 
+    private fun addTemporaryZapOptimistically(zapAction: UiEvent.ZapStream, activeAccount: UserAccount): String {
```
markocic: Could we change the name? Maybe just `addZapOptimistically`.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -614,6 +623,28 @@ class LiveStreamViewModel @AssistedInject constructor(
...
+        val zapAmount = zapAction.zapAmount ?: _state.value.zappingState.zapDefault.amount.toULong()
+        val tempZapId = "${UUID.randomUUID()}"
```
markocic: Could we just do `val tempZapId = UUID.randomUUID().toString()`? Using string interpolation to force `.toString()` call seems a bit off.

AleksandarIlic (commented):
How are we sure that we are not duplicating the zaps? The one added optimistically + the same zap that comes from a subscription?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -408,16 +416,44 @@ class LiveStreamViewModel @AssistedInject constructor(
                 setState { copy(comment = TextFieldValue(), taggedUsers = emptyList()) }
             } catch (error: NostrPublishException) {
                 Timber.w(error)
+                chatMessages = chatMessages?.filterNot { it.uniqueId == tempMessageId }
```
AleksandarIlic, line 419: Instead of `.filterNot` could we use just `remove` since we are actually removing only one element? If yes, we should change this to all places for chat messages and zaps.

AleksandarIlic (changes requested):

PR #667, opened 2025-09-22
Improve live stream playback stability

mehmedalijaK:


markocic (approved):
These changes seem good, although I am not sure they are fixing the stability issues we were having. I ran a comparison with my device running `2.5.11` and the device stream ran longer. This could be due to one being emulator and another phisical device.

PR #668, opened 2025-09-24
Fix/stream deep linking

mehmedalijaK:


data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -140,6 +144,31 @@ class StreamRepositoryImpl(
             .map { list -> list.map { it.asStreamDO() } }
             .distinctUntilChanged()
 
+    override suspend fun findStreamNaddr(hostPubkey: String, identifier: String): Result<Naddr> =
+        withContext(dispatcherProvider.io()) {
+            runCatching {
+                val response = liveStreamApi.findLiveStream(
+                    body = FindLiveStreamRequestBody(hostPubkey = hostPubkey, identifier = identifier),
+                )
+                val liveActivity = response.liveActivity ?: throw NetworkException("Live stream event not found.")
```
markocic: Not sure if we should throw `NetworkException`, this is not a network issue.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -140,6 +144,31 @@ class StreamRepositoryImpl(
...
+                val dTag = liveActivity.tags.findFirstIdentifier()
```
markocic: This check is redundant. We are returning `null` from `asStreamData()` on missing identifier tag. We could return `Result.failure` if `streamData` is `null` and `Result.success(Naddr)` otherwise.

data/caching/remote/src/commonMain/kotlin/net/primal/data/remote/api/stream/model/LiveEventsFromFollowsRequest.kt
```diff
@@ -7,3 +7,9 @@ import kotlinx.serialization.Serializable
 data class LiveEventsFromFollowsRequest(
     @SerialName("user_pubkey") val pubkey: String,
 )
+
+@Serializable
+data class FindLiveStreamRequestBody(
```
AleksandarIlic: Why is this data class inside of this file? Why not it's own file?

data/caching/remote/src/commonMain/kotlin/net/primal/data/remote/api/stream/model/LiveEventsFromFollowsResponse.kt
```diff
@@ -7,3 +7,7 @@ import net.primal.domain.nostr.NostrEvent
 data class LiveEventsFromFollowsResponse(
     val liveActivity: List<NostrEvent>,
 )
+
+data class FindLiveStreamResponse(
```
AleksandarIlic: Why here? This should be in it's own file.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -140,6 +144,31 @@ class StreamRepositoryImpl(
...
+                    ?: throw NetworkException("Identifier not found in event.")
```
AleksandarIlic: Let's avoid throwing exceptions. They are not very convenient to handle and they also cause issues when code gets compiled for iOS. We can use `Result.failure` if something goes wrong while finding the stream naddr.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/streams/StreamRepositoryImpl.kt
```diff
@@ -140,6 +141,33 @@ class StreamRepositoryImpl(
...
+                        userId = streamData.eventAuthorId,
+                        identifier = streamData.dTag,
+                    )
+                    Result.success(naddr)
+                }
+            } catch (e: Exception) {
+                Result.failure(e)
+            }
+        }
+
```
markocic, line 166: We should stick with `runCatching`.

markocic (changes requested):

PR #669, opened 2025-09-24
Prevent crash when Amber signer is uninstalled

mehmedalijaK:


app/src/main/kotlin/net/primal/android/signer/AmberLauncher.kt
```diff
@@ -128,8 +136,14 @@ fun AmberLauncher.launchGetPublicKey() {
  * @param event The [NostrUnsignedEvent] that needs to be signed.
  *
  * @see rememberAmberSignerLauncher
+ * @throws AmberNotInstalledException if Amber is not installed on the device.
  */
-fun AmberLauncher.launchSignEvent(event: NostrUnsignedEvent) {
+@Throws(AmberNotInstalledException::class)
+fun AmberLauncher.launchSignEvent(context: Context, event: NostrUnsignedEvent) {
+    if (!isCompatibleAmberVersionInstalled(context)) {
```
markocic: There is no need for manually checking each time. `launch` below will throw `ActivitiyNotFoundException` if there is no activity that can handle launched intent. We should handle that exception instead and show appropriate message.

app/src/main/kotlin/net/primal/android/auth/login/LoginScreen.kt
```diff
@@ -125,7 +127,11 @@ fun LoginScreen(
         onFailure = { eventPublisher(LoginContract.UiEvent.ResetLoginState) },
     ) { pubkey ->
         eventPublisher(LoginContract.UiEvent.UpdateLoginInput(newInput = pubkey, loginType = LoginType.ExternalSigner))
-        signLauncher.launchSignEvent(event = buildAppSpecificDataEvent(pubkey = pubkey))
+        try {
+            signLauncher.launchSignEvent(context = context, event = buildAppSpecificDataEvent(pubkey = pubkey))
+        } catch (error: AmberNotInstalledException) {
+            Toast.makeText(context, error.message, Toast.LENGTH_SHORT).show()
```
markocic: We should show custom message from `strings.xml`. This goes for other places as well.

app/src/main/kotlin/net/primal/android/messages/security/Nip04MessageCipher.kt
```diff
@@ -44,6 +48,9 @@ class Nip04MessageCipher @Inject constructor(
     ): String {
         val npub = userId.hexToNpubHrp()
         return if (credentialsStore.isExternalSignerLogin(npub = npub)) {
+            if (!isCompatibleAmberVersionInstalled(context)) {
```
markocic: Do we need this check? Can `encryptNip04WithAmber` throw some exception? If so, it would be better to handle that exception instead of checking in the first place. Same thing applies to the decrypt function as well.

markocic (changes requested):

app/src/main/kotlin/net/primal/android/signer/AmberNotInstalledException.kt
```diff
@@ -0,0 +1,3 @@
+package net.primal.android.signer
+
+class AmberNotInstalledException(message: String = "Amber Signer is not installed.") : Exception(message)
```
AleksandarIlic: We should not throw our custom exceptions. The root cause is that system is throwing `ActivitiyNotFoundException`. This is a known exception on Android and we should just handle it where applicable.

app/src/main/kotlin/net/primal/android/auth/login/LoginScreen.kt
```diff
@@ -163,7 +169,13 @@ fun LoginScreen(
             uiMode = uiMode,
             onLoginInputChanged = { eventPublisher(LoginContract.UiEvent.UpdateLoginInput(newInput = it)) },
             onLoginClick = { eventPublisher(LoginContract.UiEvent.LoginRequestEvent()) },
-            onLoginWithAmberClick = { pubkeyLauncher.launchGetPublicKey() },
+            onLoginWithAmberClick = {
+                try {
+                    pubkeyLauncher.launchGetPublicKey(context = context)
+                } catch (error: AmberNotInstalledException) {
```
AleksandarIlic: I'm not sure if this is necessary here. Aren't we on Login screen checking if Amber is installed and show this option only in that case. There should never be a situation where this gets called without installed Amber, right?

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/messages/security/Nip04MessageCipher.kt
mehmedalijaK: The encryptNip04WithAmber function does not throw an exception when the Amber app is not installed; instead, it returns null. This null result is ambiguous because it also occurs if the user simply cancels the operation within Amber.

The proactive isCompatibleAmberVersionInstalled check allows us to distinguish between these two different failure scenarios.

app/src/main/kotlin/net/primal/android/signer/AmberLauncher.kt
```diff
@@ -92,8 +95,14 @@ fun rememberAmberSignerLauncher(onFailure: ((ActivityResult) -> Unit)? = null, o
  * errors in the flow), use it together with [rememberAmberPubkeyLauncher].
  *
  * @see rememberAmberPubkeyLauncher
+ * @throws ActivityNotFoundException if Amber is not installed on the device.
  */
-fun AmberLauncher.launchGetPublicKey() {
+@Throws(ActivityNotFoundException::class)
+fun AmberLauncher.launchGetPublicKey(context: Context) {
+    if (!isCompatibleAmberVersionInstalled(context)) {
```
markocic: The `launch` below throws `ActivityNotFoundException` if there is no activity that can handle our intent. No need to check for it manually. Remove this.

app/src/main/kotlin/net/primal/android/messages/security/Nip04MessageCipher.kt
markocic: We are using `ContentResolver` to encrypt and decrypt messages. As such, operation happens in the background and user doesn't actually confirm/cancel the operation within Amber. I don't think we should introduce dependency to `ApplicationContext` for the sole purpose of knowing what exactly happened when at the end of the day we are throwing same exception in case Amber fails to complete the operation.

Revert this file and confirm that the app doesn't crash if Amber is missing.

markocic (approved):

PR #673, opened 2025-09-29
Implement Delete Reposts dialog for notes

markocic:


data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/feed/FeedRepositoryImpl.kt
```diff
@@ -73,6 +75,36 @@ internal class FeedRepositoryImpl(
...
+            database.reposts().findByPostId(postId = postId, authorId = userId)?.let { repostData ->
+                Result.success(
+                    FeedPostRepostInfo(
+                        repostId = repostData.repostId,
+                        repostAuthorId = repostData.authorId,
+                        repostAuthorDisplayName = null,
+                        repostCreatedAt = repostData.createdAt,
+                    ),
+                )
+            } ?: Result.failure(IllegalArgumentException("User with given userId didn't repost target post."))
```
AleksandarIlic: This is not an Illegal Argument Exception. The argument is fine, it's just that we could not find the data for those arguments.

app/src/main/kotlin/net/primal/android/notes/feed/NoteRepostOrQuoteBottomSheet.kt
```diff
@@ -1,32 +1,43 @@
...
 import net.primal.android.core.compose.IconText
 import net.primal.android.core.compose.button.PrimalFilledButton
+import net.primal.android.core.compose.foundation.isAppInDarkPrimalTheme
 import net.primal.android.core.compose.icons.PrimalIcons
+import net.primal.android.core.compose.icons.primaliconpack.DeleteRepost
 import net.primal.android.core.compose.icons.primaliconpack.Quote
 import net.primal.android.core.compose.icons.primaliconpack.Repost
 import net.primal.android.theme.AppTheme
 
+private val DELETE_REPOST_COLOR
```
AleksandarIlic: We should try to reuse `AppTheme.colorScheme.error` if the colors are similar.

PR #689, opened 2025-10-07
Implement Stream Ended support

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -489,11 +491,11 @@ class LiveStreamViewModel @AssistedInject constructor(
                     setState {
                         copy(
                             streamInfoLoading = false,
-                            isStreamUnavailable = streamUrlToPlay == null,
+                            isStreamUnavailable = streamUrlToPlay == null && !isEnded,
                             playerState = playerState.copy(
                                 isLive = isLive,
                                 atLiveEdge = isLive,
-                                isVideoFinished = false,
+                                isVideoFinished = if (isEnded) true else playerState.isVideoFinished,
```
markocic: This can be simplified to `isEnded || playerState.isVideoFinished`.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayerBox.kt
```diff
@@ -69,40 +71,63 @@ fun LiveStreamPlayerBox(
...
+private fun StreamUnavailableContent(
+    modifier: Modifier,
+    state: LiveStreamContract.UiState,
+    onRetryClick: (() -> Unit)?,
+) {
+    Box(
+        modifier = modifier
+            .fillMaxSize()
+            .background(AppTheme.extraColorScheme.surfaceVariantAlt1),
+        contentAlignment = Alignment.Center,
```
markocic: There is no need for this outer `Box` composable. We should pay attention to extra elements as they decrease performance over time.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayerBox.kt
```diff
@@ -69,40 +71,63 @@ fun LiveStreamPlayerBox(
...
+                Text(
+                    text = stringResource(id = R.string.live_stream_recording_not_available).uppercase(),
+                    color = AppTheme.extraColorScheme.onSurfaceVariantAlt4,
+                    style = AppTheme.typography.bodyLarge.copy(
+                        fontSize = 16.sp,
+                        lineHeight = 20.sp,
+                        fontWeight = FontWeight.Bold,
+                    ),
+                    textAlign = TextAlign.Center,
+                )
```
markocic: Only thing that is different here is the `text`. We should use single `Text` here and just change the displaying text to avoid keeping both of them in sync.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayerBox.kt
```diff
@@ -69,40 +71,63 @@ fun LiveStreamPlayerBox(
...
-                            IconButton(onClick = onRetryClick) {
-                                ShadowIcon(
-                                    modifier = Modifier.size(30.dp),
-                                    tint = Color.White,
-                                    imageVector = Icons.Default.Refresh,
-                                    contentDescription = stringResource(id = R.string.live_stream_retry_button),
-                                )
-                            }
+@Composable
+private fun StreamUnavailableContent(
```
markocic: Let's go with `StreamFallbackContent`. I don't like the word `Unavailable` here as it is one of the states and can be misleading.

PR #691, opened 2025-10-08
Implement replay button on stream and recording ended state

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/ui/ExpandedLiveStreamPlayer.kt
```diff
@@ -81,6 +83,7 @@ fun ExpandedLiveStreamPlayer(
 
             if (streamUrl != null) {
                 mediaController.setMediaItem(buildMediaItem(state.naddr, streamUrl, state.streamInfo))
+                mediaController.repeatMode = Player.REPEAT_MODE_OFF
```
markocic: We have similar logic in `LiveStreamMiniPlayer`, should we add this there too?

app/src/main/kotlin/net/primal/android/stream/LiveStreamOverlay.kt
```diff
@@ -207,6 +207,11 @@ private fun LiveStreamAnimatedContent(
                         callbacks = callbacks,
                         sharedTransitionScope = this@SharedTransitionLayout,
                         animatedVisibilityScope = this,
+                        onReplay = {
+                            mediaController.seekTo(0L)
+                            mediaController.play()
+                            viewModel.setEvent(UiEvent.OnReplayStream)
+                        },
```
markocic: I think we should lose `onReplay` and `onRetry` arguments here. We are already forwarding `mediaController` and `eventPublisher` which is all that is needed for these callbacks. I guess we missed this with `onRetry` before.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -355,6 +358,18 @@ class LiveStreamViewModel @AssistedInject constructor(
             }
         }
 
+    private fun replayStream() {
+        val recordingUrl = state.value.streamInfo?.recordingUrl ?: return
+        viewModelScope.launch {
+            setState {
+                copy(
+                    playerState = playerState.copy(isVideoFinished = false),
+                    streamInfo = this.streamInfo?.copy(streamUrl = recordingUrl),
```
markocic: This looks a little weird. I thought `streamInfo.streamUrl` represents `streamingUrl` and `recordingUrl` represented just that and we are just switching based on state. This could potentially lead to unexpected behaviour along the road.

Reading `LiveStreamContract` should give us a good idea of what the state represents which in this case leads a different path. We might need to change logic or use better naming. I am thinking former mentioned here. `StreamInfo` should represent `Stream` which demands logic change.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -526,6 +552,19 @@ class LiveStreamViewModel @AssistedInject constructor(
...
+        isEnded: Boolean,
+        currentStreamInfo: StreamInfoUi?,
+        stream: Stream,
+    ): String? {
+        return when {
+            currentStreamInfo == null -> if (isLive) stream.streamingUrl else stream.recordingUrl
+            currentStreamInfo.streamStatus == StreamStatus.LIVE && isEnded -> null
+            else -> currentStreamInfo.streamUrl
+        }
+    }
```
markocic: With the mentioned logic change I think this function will be improved. Just looking at the function arguments I had follow up questions. `StreamInfoUi` and `Stream` should represent the same thing and needing both is the indication that we are not doing something right or setting ourselves up for failure.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayerBox.kt
```diff
@@ -108,16 +126,34 @@ private fun StreamFallbackContent(
...
-                }
+        if (isFinished && hasRecording && onReplayClick != null) {
+            Button(
+                modifier = Modifier.height(26.dp),
+                onClick = onReplayClick,
+                shape = AppTheme.shapes.extraLarge,
+                colors = ButtonDefaults.buttonColors(
+                    containerColor = Color.White,
+                    contentColor = Color.Black,
+                ),
```
markocic, line 145: These colors won't work for light theme. Also, they aren't full black and white. They are a shade brighter/darker. Check color codes in the design.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamMiniPlayer.kt
```diff
@@ -111,13 +113,14 @@ fun LiveStreamMiniPlayer(
...
+        val newStreamUrl = state.playbackUrl
         if (newStreamUrl != null) {
             val currentMediaItem = mediaController.currentMediaItem
             val currentMediaItemUri = currentMediaItem?.localConfiguration?.uri?.toString()
             if (newStreamUrl != currentMediaItemUri) {
                 mediaController.setMediaItem(buildMediaItem(state.naddr, newStreamUrl, state.streamInfo))
+                mediaController.repeatMode = Player.REPEAT_MODE_OFF
                 mediaController.prepare()
                 mediaController.playWhenReady = true
             }
```
markocic: I think this logic is flawed. If `playbackUrl` suddenly becomes `null`, the stream won't be stopped. Right now this _might_ not be an issue but it's a possibility. Let's use logic from `ExpandedLiveStreamPlayer`. Extract that into some helper composable that will run this `LaunchedEffect` and use it there and here. That way we will centralize logic in one place.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayerBox.kt
```diff
@@ -74,18 +80,51 @@ fun LiveStreamPlayerBox(
...
+        Color(0xFF121212)
+    }
+
+private val ReplayButtonContentColor: Color
+    @Composable
+    get() = if (LocalPrimalTheme.current.isDarkTheme) {
+        Color(0xFF121212)
+    } else {
+        Color(0xFFFFFFFF)
+    }
```
markocic: Light theme should be `F5F5F5`. We have this color in our theme. `AppTheme.extraColorScheme.surfaceVariantAlt2`.

app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamPlayerBox.kt
```diff
@@ -74,18 +80,51 @@ fun LiveStreamPlayerBox(
...
     }
 }
 
+private val ReplayButtonContainerColor: Color
+    @Composable
+    get() = if (LocalPrimalTheme.current.isDarkTheme) {
+        Color(0xFFFFFFFF)
+    } else {
+        Color(0xFF121212)
+    }
```
markocic: We have this color in our theme. `AppTheme.colorScheme.onSurface`.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -355,6 +357,18 @@ class LiveStreamViewModel @AssistedInject constructor(
             }
         }
 
+    private fun replayStream() {
+        val recordingUrl = state.value.streamInfo?.recordingUrl ?: return
```
markocic, line 362: Is it possible for us to display Replay button with `recordingUrl` being `null`?

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -526,6 +553,21 @@ class LiveStreamViewModel @AssistedInject constructor(
...
+        newStreamingUrl: String?,
+        newRecordingUrl: String?,
+    ): String? {
+        return when {
+            currentStreamInfo == null -> if (isLive) newStreamingUrl else newRecordingUrl
+            currentStreamInfo.streamStatus == StreamStatus.LIVE && isEnded -> null
+            else -> currentPlaybackUrl
+        }
+    }
+
```
markocic: I understand the problem with this function. Root cause was hidden in `StreamInfoUi`. I believe it is flawed for having `ProfileDetailsUi` and `ProfileStatsUi` as they are changed in another place. Let's ignore this for now as that would require further changes and we should focus on merging this for now.

Since we are now dealing with too much arguments, let's just use two.
`currentStreamInfo: StreamInfoUi?` and `updatedStreamInfo: Stream`.
Rest of the arguments can be constructed from these two. Ideally we should have both be `StreamInfoUi` but that is not possible right now.

app/src/main/kotlin/net/primal/android/stream/LiveStreamViewModel.kt
```diff
@@ -482,27 +496,40 @@ class LiveStreamViewModel @AssistedInject constructor(
...
 
                     setState {
                         copy(
                             streamInfoLoading = false,
-                            isStreamUnavailable = streamUrlToPlay == null && !isEnded,
+                            isStreamUnavailable = nextPlaybackUrl == null && !isEnded,
+                            playbackUrl = nextPlaybackUrl,
                             playerState = playerState.copy(
                                 isLive = isLive,
                                 atLiveEdge = isLive,
```
markocic, line 521: Side note: I just realized this could lead to a bug.

Scenario: User watches the stream but is a bit behind (not `atLiveEdge`). He is watching in Mini Player and scrolling through his feed. He sees someone posting about the stream he is watching and stream is updated in the db. This causes this observe logic to trigger and places him `atLiveEdge` even though he is not.

Ignore this comment for now. We should fix this in another PR.

app/src/main/kotlin/net/primal/android/stream/ui/StreamPlaybackController.kt
```diff
@@ -0,0 +1,34 @@
...
+import net.primal.domain.nostr.Naddr
+
+@Composable
+fun StreamPlaybackController(
+    mediaController: Player,
+    playbackUrl: String?,
+    naddr: Naddr?,
+    streamInfo: LiveStreamContract.StreamInfoUi?,
+) {
+    LaunchedEffect(playbackUrl) {
```
markocic: We should have `mediaController` and `playbackUrl` as keys.

PR #695, opened 2025-10-10
Fix zap targets with recipient user id and lnUrl

AleksandarIlic:
When zapping in `ZapRequest`, we were using the event author’s pubkey instead of the recipient’s pubkey. This PR:
- adds `recipientUserId` and `recipientLnUrlDecoded` to  `ZapTarget` entity as mandatory (and common) fields for all targets;
- adds specific `naddr` in `ZapTarget.ReplaceableEvent`; and
- adjusts the codebase accordingly to the changes.

The issue was found when zapping a stream. We were using the event author's pubkey and recipient's lnUrl; and now with this PR recipient data should be aligned.

### DIFF
--- a/domain/nostr/src/commonMain/kotlin/net/primal/domain/nostr/zaps/ZapTarget.kt
+++ b/domain/nostr/src/commonMain/kotlin/net/primal/domain/nostr/zaps/ZapTarget.kt
@@ -27,7 +27,7 @@ sealed class ZapTarget(
     )
 
     data class ReplaceableEvent(
-        val naddr: String,
+        val aTag: String,
         val eventId: String,
         override val recipientUserId: String,
         override val recipientLnUrlDecoded: String,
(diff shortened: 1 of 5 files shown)


domain/nostr/src/commonMain/kotlin/net/primal/domain/nostr/zaps/ZapTarget.kt
```diff
@@ -5,58 +5,51 @@ import net.primal.domain.nostr.asEventIdTag
...
+        override val recipientLnUrlDecoded: String,
+    ) : ZapTarget(
+        recipientUserId = recipientUserId,
+        recipientLnUrlDecoded = recipientLnUrlDecoded,
+    )
 
     data class ReplaceableEvent(
-        val kind: Int,
-        val identifier: String,
+        val naddr: String,
```
markocic: This should be called `aTag`.

domain/primal/src/commonMain/kotlin/net/primal/domain/feeds/DvmFeed.kt
```diff
@@ -15,7 +17,9 @@ data class DvmFeed(
     val kind: FeedSpecKind? = null,
     val isPrimalFeed: Boolean? = null,
     val actionUserIds: List<String> = emptyList(),
-)
+) {
+    val dvmNaddr = "${NostrEventKind.AppHandler.value}:${this.dvmPubkey}:${this.dvmId}"
```
markocic: This should be `dvmATag`.

PR #697, opened 2025-10-13
Prevent data fetches with empty pubkey on login

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/home/HomeFeedViewModel.kt
```diff
@@ -114,7 +115,8 @@ class HomeFeedViewModel @Inject constructor(
     @OptIn(FlowPreview::class)
     private fun observeLiveEventsFromFollows() =
         viewModelScope.launch {
-            streamRepository.observeLiveEventsFromFollows(userId = activeAccountStore.activeUserId())
+            val userId = activeAccountStore.activeUserId.first { it.isNotEmpty() }
```
markocic: Not sure about `first` here. `first` can throw `NoSuchElementException`. We should opt for a non terminating option.

app/src/main/kotlin/net/primal/android/core/updater/DataUpdater.kt
```diff
@@ -22,15 +25,18 @@ class DataUpdater @Inject constructor(
...
+    private val userDataUpdater = _userDataUpdater.asStateFlow()
 
     init {
         observeActiveAccount()
     }
 
     fun updateData() =
         scope.launch {
-            userDataUpdater?.updateWithDebounce(30.minutes)
+            val updater = userDataUpdater.first { it != null }
```
markocic: This would throw `NoSuchElementException` in case `activeUserId` comes in as blank and `updateData` gets called.

app/src/main/kotlin/net/primal/android/core/updater/DataUpdater.kt
```diff
@@ -22,15 +25,18 @@ class DataUpdater @Inject constructor(
     private val updateStaleStreamDataUseCase: UpdateStaleStreamDataUseCase,
 ) {
     private val scope = CoroutineScope(dispatcherProvider.io() + SupervisorJob())
-    private var userDataUpdater: UserDataUpdater? = null
+    private val _userDataUpdater = MutableStateFlow<UserDataUpdater?>(null)
+    private val userDataUpdater = _userDataUpdater.asStateFlow()
```
markocic: Why is this `MutableStateFlow`? I might be missing something but I can't see the reason for it.

app/src/main/kotlin/net/primal/android/notes/home/HomeFeedViewModel.kt
```diff
@@ -135,7 +137,7 @@ class HomeFeedViewModel @Inject constructor(
         viewModelScope.launch {
             try {
                 setState { copy(loading = true) }
-                val userId = activeAccountStore.activeUserId()
+                val userId = activeAccountStore.activeUserId.first { it.isNotEmpty() }
```
AleksandarIlic: I don't like this approach. I think we are using `activeAccountStore.activeUserId` like this on lot of screens and it works fine. Or it might be `activeAccountStore.activeUserId()`, not sure now.

PR #698, opened 2025-10-14
Prevent click-through when dismissing stream chat keyboard

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/ui/LiveStreamScreen.kt
```diff
@@ -601,39 +603,53 @@ private fun LiveStreamContent(
...
+                                Box(
+                                    modifier = Modifier
+                                        .fillMaxSize()
+                                        .clickable(
+                                            interactionSource = remember { MutableInteractionSource() },
+                                            indication = null,
+                                            onClick = { focusManager.clearFocus() },
+                                        ),
+                                )
+                            }
```
markocic, line 652: Im not sure about this approach. This will permanently add another layer which affects performance. Could we just change `onClick` events of chat message items to hide keyboard if shown and open bottom sheet otherwise?

PR #704, opened 2025-10-16
Implement complete user data cleanup on logout

mehmedalijaK:


data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/repository/WalletRepositoryImpl.kt
```diff
@@ -126,6 +126,24 @@ internal class WalletRepositoryImpl(
...
+    override suspend fun deleteAllUserData(userId: String) =
+        withContext(dispatcherProvider.io()) {
+            walletDatabase.withTransaction {
+                val wallets = walletDatabase.wallet().findWalletInfosByUserId(userId.asEncryptable())
+                val walletIds = wallets.map { it.walletId }
+
+                if (walletIds.isNotEmpty()) {
+                    walletDatabase.walletSettings().deleteWalletSettings(walletIds)
+                    walletIds.forEach { walletId ->
+                        walletDatabase.wallet().deleteWalletById(walletId)
```
AleksandarIlic: Dao could accept list of Ids and then you can delete all wallets with single database operation.

data/wallet/local/src/commonMain/kotlin/net/primal/wallet/data/local/dao/WalletTransactionDao.kt
```diff
@@ -41,4 +42,23 @@ interface WalletTransactionDao {
...
+    suspend fun deleteNostrTransactionsByIds(transactionIds: List<String>)
+
+    @Transaction
+    suspend fun deleteAllUserTransactions(userId: Encryptable<String>) {
+        val txIds = findTransactionIdsByUserId(userId)
+        if (txIds.isNotEmpty()) {
+            deletePrimalTransactionsByIds(txIds)
+            deleteNostrTransactionsByIds(txIds)
+        }
+        deleteAllTransactionsByUserId(userId)
```
AleksandarIlic: Can `deleteAllTransactionsByUserId` be private function? It can be confused with `deleteAllUserTransactions`. Also, `deletePrimalTransactionsByIds` and `deleteNostrTransactionsByIds` could private as well if they are not used in other places.

data/wallet/local/src/commonMain/kotlin/net/primal/wallet/data/local/dao/WalletDao.kt
```diff
@@ -31,9 +31,13 @@ interface WalletDao {
     @Insert(onConflict = OnConflictStrategy.IGNORE)
     suspend fun insertOrIgnoreWalletInfo(info: WalletInfo)
 
+    @Transaction
     @Query("SELECT * FROM WalletInfo WHERE userId = :userId")
     fun observeWalletsByUserId(userId: Encryptable<String>): Flow<List<Wallet>>
 
+    @Query("SELECT * FROM WalletInfo WHERE userId = :userId")
+    suspend fun findWalletInfosByUserId(userId: Encryptable<String>): List<WalletInfo>
```
AleksandarIlic: Let's put this function next to the `findWalletInfo` since they are dealing with `WalletInfo`.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -161,7 +161,7 @@ class WalletSettingsViewModel @AssistedInject constructor(
                 walletId = state.value.wallet?.walletId ?: activeAccountStore.activeUserId(),
                 spamThresholdAmountInSats = amountInSats,
             )
-            walletRepository.deleteAllTransactions(userId = activeAccountStore.activeUserId())
+            walletRepository.deleteAllUserData(userId = activeAccountStore.activeUserId())
```
AleksandarIlic: Uh, uh... I think we should not do this. This function is changing the spam threshold amount. That change requires that all transaction be wiped, but the wallet account should remain on the device. What you need to do here is to ONLY delete all transactions for the `userId`.

domain/wallet/src/commonMain/kotlin/net/primal/domain/wallet/WalletRepository.kt
```diff
@@ -27,7 +27,7 @@ interface WalletRepository {
 
     suspend fun findTransactionByIdOrNull(txId: String): Transaction?
 
-    suspend fun deleteAllTransactions(userId: String)
+    suspend fun deleteAllUserData(userId: String)
```
AleksandarIlic, line 32: You are gonna need to re-add `deleteAllTransactions` because it's needed when changing spam threshold. I was not aware of that when I ask you to remove it. Sorry.

PR #705, opened 2025-10-16
Correct Started now ago timestamp display

mehmedalijaK:


app/src/main/kotlin/net/primal/android/stream/ui/StreamMetaData.kt
```diff
@@ -42,11 +42,14 @@ fun StreamMetaData(
         StreamLiveIndicator(isLive = isLive, textColor = textColor, fontSize = 15.sp, lineHeight = 16.sp)
 
         if (startedAt != null) {
+            val timeAgo = Instant.ofEpochSecond(startedAt).asBeforeNowFormat()
+            val startedAtText = if (timeAgo.equals("now", ignoreCase = true)) {
```
markocic: This wouldn't work if we change the language sometime in the future.

markocic (changes requested):

PR #710, opened 2025-10-20
Prevent zap cancellation on screen exit

mehmedalijaK:


app/src/main/kotlin/net/primal/android/wallet/zaps/ZapHandler.kt
```diff
@@ -29,33 +32,38 @@ class ZapHandler @Inject constructor(
...
-        eventInteractionRepository.zapEvent(
-            userId = userId,
-            walletId = walletId,
-            amountInSats = zapAmountInSats,
-            comment = zapComment,
-            target = target,
-            zapRequestEvent = userZapRequestEvent,
-        )
-    }
+                eventInteractionRepository.zapEvent(
```
AleksandarIlic: This is the most time consuming part of the zapping. We should wrap only this block with `NonCancellable` and not the entire zap function.

app/src/main/kotlin/net/primal/android/wallet/zaps/ZapHandler.kt
```diff
@@ -29,33 +32,38 @@ class ZapHandler @Inject constructor(
...
+                    userId = userId,
+                    walletId = walletId,
+                    amountInSats = zapAmountInSats,
+                    comment = zapComment,
+                    target = target,
+                    zapRequestEvent = userZapRequestEvent,
+                )
+            } ?: ZapResult.Failure(error = ZapError.Unknown())
```
AleksandarIlic: We should add `ZapError.Timeout()` since it's a known error. On this UI level this should be processed along with `ZapError.FailedToPublishEvent` and `ZapError.FailedToSignEvent`.

app/src/main/kotlin/net/primal/android/wallet/zaps/ZapHandler.kt
```diff
@@ -49,13 +53,28 @@ class ZapHandler @Inject constructor(
...
-            userId = userId,
-            walletId = walletId,
-            amountInSats = zapAmountInSats,
-            comment = zapComment,
-            target = target,
-            zapRequestEvent = userZapRequestEvent,
-        )
+        runCatching {
+            withContext(NonCancellable) {
+                withTimeout(30.seconds) {
```
markocic, line 58: This won't actually run for only `30.seconds`.
`withTimeout` will throw `TimeoutCancellationException : CancellationException` but `NonCancellable` ignores all `CancellationException`s thrown inside this context.

In worst case scenario, this block will hang for uncertain amount of time.

Solution is to wrap `withTimeout` with a new job that is cancellable so when the timeout runs out, that job will be cancelled but outside cancellation won't affect newly created job.

AleksandarIlic: I'm not sure about this. I just run a simple test code and it seems that it gets properly canceled on timeout.

markocic: Guess I misunderstood the docs. Thanks for catching this! Approved.

markocic (dismissed):

PR #716, opened 2025-10-22
Implement local account switcher in Note Editor

mehmedalijaK:


app/src/main/kotlin/net/primal/android/editor/NoteEditorContract.kt
```diff
@@ -53,6 +57,9 @@ interface NoteEditorContract {
         data class SearchUsers(val query: String) : UiEvent()
         data class ToggleSearchUsers(val enabled: Boolean) : UiEvent()
         data class TagUser(val taggedUser: NoteTaggedUser) : UiEvent()
+        data object ShowAccountSwitcher : UiEvent()
+        data object HideAccountSwitcher : UiEvent()
+        data class AccountSelected(val accountId: String) : UiEvent()
```
markocic: nitpick: We should keep event names in present tense. This would be `SelectAccount`.

app/src/main/kotlin/net/primal/android/editor/NoteEditorContract.kt
```diff
@@ -34,6 +35,9 @@ interface NoteEditorContract {
         val taggedUsers: List<NoteTaggedUser> = emptyList(),
         val referencedNostrUris: List<ReferencedUri<*>> = emptyList(),
         val userTaggingState: UserTaggingState = UserTaggingState(),
+        val showAccountSwitcher: Boolean = false,
+        val availableAccounts: List<UserAccountUi> = emptyList(),
+        val selectedAccountId: String? = null,
```
markocic: As far as I can see, `selectedAccountId` is connected with `activeAccountAvatarCdnImage`, `activeAccountLegendaryCustomization` and `activeAccountBlossoms`. This is not obvious until you read code in `NoteEditorViewModel`.

Let's have a single field `selectedAccount: UserAccountUi? = null`. On init it should be current active account.

app/src/main/kotlin/net/primal/android/editor/NoteEditorViewModel.kt
```diff
@@ -265,10 +265,46 @@ class NoteEditorViewModel @AssistedInject constructor(
...
+                    setState {
+                        copy(
+                            activeAccountAvatarCdnImage = selectedAccount.avatarCdnImage,
+                            activeAccountLegendaryCustomization = selectedAccount.legendaryCustomization,
+                            activeAccountBlossoms = selectedAccount.avatarBlossoms,
+                        )
+                    }
+                }
+        }
+    }
```
markocic: Let's lose all this. This is just reacting to changes in `selectedAccountId` and finding corresponding account to set the state. We can just set correct account immediately.

app/src/main/kotlin/net/primal/android/drawer/multiaccount/ui/AccountSwitcherBottomSheet.kt
```diff
@@ -59,11 +59,13 @@ fun AccountSwitcherBottomSheet(
...
-    onLogoutClick: (String) -> Unit,
     onAccountClick: (String) -> Unit,
-    onCreateNewAccountClick: () -> Unit,
-    onAddExistingAccountClick: () -> Unit,
     sheetState: SheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true),
+    onLogoutClick: (String) -> Unit = {},
+    onCreateNewAccountClick: () -> Unit = {},
+    onAddExistingAccountClick: () -> Unit = {},
+    showEditButton: Boolean = true,
+    showAddAccountButtons: Boolean = true,
```
markocic: Let's have all callbacks nullable with `null` as default. If we have a `null` callback, we won't show button corresponding to it.

app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
```diff
@@ -135,6 +136,26 @@ fun NoteEditorScreen(
     val snackbarHostState = remember { SnackbarHostState() }
     val context = LocalContext.current
 
+    if (state.showAccountSwitcher) {
+        val activeAccount = remember(state.selectedAccountId, state.availableAccounts) {
+            state.availableAccounts.find { it.pubkey == state.selectedAccountId }
```
markocic: Side note (this should be gone after fixing the rest of comments): Conditionally calling this makes `remember` "less effective". It is only ran after we want to show switcher but is removed when we close it.

Take a look at the note at the end of first paragraph:
https://developer.android.com/develop/ui/compose/state#state-in-composables

app/src/main/kotlin/net/primal/android/editor/NoteEditorContract.kt
```diff
@@ -26,14 +25,14 @@ interface NoteEditorContract {
...
-        val activeAccountAvatarCdnImage: CdnImage? = null,
-        val activeAccountLegendaryCustomization: LegendaryCustomization? = null,
-        val activeAccountBlossoms: List<String> = emptyList(),
+        val selectedAccount: UserAccountUi? = null,
         val uploadingAttachments: Boolean = false,
         val attachments: List<NoteAttachment> = emptyList(),
         val taggedUsers: List<NoteTaggedUser> = emptyList(),
         val referencedNostrUris: List<ReferencedUri<*>> = emptyList(),
         val userTaggingState: UserTaggingState = UserTaggingState(),
+        val showAccountSwitcher: Boolean = false,
```
markocic: Not sure about keeping this in the state. It can go to UI as we are not doing any computations when modifying sheet visibility.

AleksandarIlic: Great catch. I agree.

PR #718, opened 2025-10-23
Implement NIP-92 for dimensions

mehmedalijaK:


data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/mappers/remote/EventUriParsers.kt
```diff
@@ -24,14 +28,44 @@ fun List<PostData>.flatMapPostsAsEventUriPO(
...
+                    url = uri,
+                    type = type,
+                    mimeType = mimeType,
+                    variants = variants + (linkThumbnailCdnResource?.variants ?: emptyList()),
+                    title = linkPreview?.title?.ifBlank { null },
+                    description = linkPreview?.description?.ifBlank { null },
+                    thumbnail = linkPreview?.thumbnailUrl?.ifBlank { null } ?: videoThumbnail,
+                    authorAvatarUrl = linkPreview?.authorAvatarUrl?.ifBlank { null },
+                )
+            }
```
markocic, line 39: Might be nitpick: This logic seems confusing. Can we split this into helper functions? From the task I can see that we should use primal cdn resource but if that is missing, we should refer to imeta tags. We can have a helper function that tries to get variants from cdn resources and then it should try to parse imeta tags. This does both simultaneously which makes understanding of the full process harder.

markocic (changes requested):

AleksandarIlic (changes requested):
NIP-92 values should not be persisted as `CdnResourceVariant` in `EventUri`. CDN variants is coupled with Primal's CDN server, while NIP-92 is nostr protocol friendly. We should keep NIP-92 info as fields in `EventUri` (or one bigger data field) and then on the UI do the resolving - we can look for CDN and if it's missing fallback to nip92 values, or whoever seems the best later.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/attachment/NoteAttachmentsExt.kt
```diff
@@ -32,3 +54,25 @@ fun CdnResourceVariant?.findImageSize(maxWidth: Int): DpSize {
...
+        width = maxWidth.dp,
+        height = finalHeight.dp,
+    )
+}
+
+private fun calculateFallbackDimensions(maxWidth: Int, type: EventUriType): DpSize {
+    return if (type == EventUriType.Video) {
+        DpSize(width = maxWidth.dp, height = (maxWidth * 9 / 16).dp)
+    } else {
+        DpSize(width = maxWidth.dp, height = maxWidth.dp)
```
markocic: Previous fallback size was `min(maxWidth, maxHeight)`. Might be good to keep it that way so we don't introduce a bug.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/attachment/NoteAttachmentsExt.kt
```diff
@@ -32,3 +54,25 @@ fun CdnResourceVariant?.findImageSize(maxWidth: Int): DpSize {
...
     )
 }
+
+private fun calculateDimensions(
+    width: Int,
+    height: Int,
+    maxWidth: Int,
+    maxHeight: Int,
+): DpSize {
+    val heightForMaxWidth = (maxWidth.toFloat() * height.toFloat()) / width.toFloat()
```
markocic: This logic is a copy-paste of the one in `CdnResourceVariantExt:20`. Could we use this function there or something like that?

PR #723, opened 2025-10-27
Implement connection flow for NostrConnect 

mehmedalijaK:


app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+                    userPubKey = selectedAccount.pubkey,
+                    connectionUrl = connectionUrl,
+                )
+                setEffect(NostrConnectContract.SideEffect.ConnectionSuccess)
+            } catch (e: Exception) {
+                Timber.e(e)
+                setEffect(NostrConnectContract.SideEffect.ConnectionFailed(e))
+            } finally {
+                setState { copy(connecting = false) }
+            }
```
markocic: Use `runCatching` instead of catching `e: Exception`.

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -553,6 +571,28 @@ private fun PrimalAppNavigation(
...
+                    type = NavType.StringType
+                    nullable = true
+                },
+                navArgument(NOSTR_CONNECT_IMAGE_URL) {
+                    type = NavType.StringType
+                    nullable = true
+                },
+            ),
+            navController = navController,
+        )
```
markocic, line 572: I think we should pass the whole url here and the dialog should do the parsing of arguments. This will make it easier to integrate scanning into other places where we have a QR code scanner.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectContract.kt
```diff
@@ -0,0 +1,54 @@
...
+    sealed class UiEvent {
+        data class TabChanged(val tab: Tab) : UiEvent()
+        data class AccountSelected(val pubkey: String) : UiEvent()
+        data class TrustLevelSelected(val level: TrustLevel) : UiEvent()
+        data object DailyBudgetClicked : UiEvent()
+        data class DailyBudgetChanged(val budget: Long?) : UiEvent()
+        data object DailyBudgetApplied : UiEvent()
+        data object DailyBudgetCancelled : UiEvent()
+        data object ConnectClicked : UiEvent()
+        data object CancelClicked : UiEvent()
```
markocic: `UiEvent`s should be in present tense.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectContract.kt
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    sealed class SideEffect {
+        data object ConnectionSuccess : SideEffect()
+        data class ConnectionFailed(val error: Throwable) : SideEffect()
+    }
+
+    companion object {
+        val DAILY_BUDGET_OPTIONS = listOf(0L, 1000L, 5000L, 10_000L, 20_000L, 50_000L, 100_000L)
+        val DAILY_BUDGET_PICKER_OPTIONS = DAILY_BUDGET_OPTIONS + listOf(null)
```
markocic, line 54: Why are these separated?

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    private fun connect() {
+        viewModelScope.launch {
+            setState { copy(connecting = true) }
+            try {
+                val selectedAccount = state.value.selectedAccount
+                    ?: throw IllegalStateException("No account selected.")
+                val connectionUrl = state.value.appUrl
+                    ?: throw IllegalStateException("No connection URL found.")
```
markocic: Maybe silent fail is a better option here. Can this even occur?

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+
+                initialize(
+                    signerPubKey = selectedAccount.pubkey,
```
markocic: Side note: We should use real `signerPubKey` for this instead of selected account.

app/src/main/kotlin/net/primal/android/redeem/RedeemCodeScreen.kt
```diff
@@ -102,7 +105,13 @@ fun RedeemCodeScreen(
...
                             modifier = Modifier.padding(bottom = 64.dp),
                             onQrCodeDetected = { eventPublisher(UiEvent.QrCodeDetected(it)) },
-                            onEnterCodeClick = { eventPublisher(UiEvent.GoToEnterCodeStage) },
+                            onEnterCodeClick = {
+                                callbacks.onNostrConnectRequest(
+                                    "Following Space",
+                                    "https://following.space/",
+                                    "https://primal.net/assets/favicon-51789dff.ico",
+                                )
+                            },
```
markocic: Just so it won't be forgotten. This change should be reversed.

app/src/main/kotlin/net/primal/android/redeem/utils/NostrConnectUrlUtils.kt.kt
```diff
@@ -0,0 +1,24 @@
...
+    return this.startsWith("$NOSTR_CONNECT_SCHEME://", ignoreCase = true)
+}
+
+private fun String.getNostrConnectQueryParameter(key: String): String? {
+    if (!this.isNostrConnectUrl()) return null
+    return try {
+        this.toUri().getQueryParameter(key)
+    } catch (_: Exception) {
+        null
+    }
```
markocic: `runCatching { ... }.getOrNull()`. We should avoid catching `Exception`. If we are doing that, it's a clear indication that we should use `runCatching`.

app/src/main/kotlin/net/primal/android/scanner/domain/QrCodeDataType.kt
```diff
@@ -1,5 +1,6 @@
 package net.primal.android.scanner.domain
 
+import net.primal.android.redeem.utils.isNostrConnectUrl // <-- DODAJ OVAJ IMPORT
```
markocic: Remove this comment.

app/src/main/kotlin/net/primal/android/redeem/utils/NostrConnectUrlUtils.kt.kt
```diff
@@ -0,0 +1,24 @@
+package net.primal.android.redeem.utils
```
markocic: Name of this file is weird.

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -1303,6 +1351,23 @@ private fun NavGraphBuilder.home(
     )
 }
 
+@OptIn(ExperimentalMaterial3Api::class)
+private fun NavGraphBuilder.nostrConnectDialog(
+    route: String,
+    arguments: List<NamedNavArgument>,
+    navController: NavController,
+) {
+    dialog(
```
markocic, line 1334: nitpick: Can we get `val viewModel = ...` and forward it as argument? For some reason we are doing that everywhere and we should keep it that way.

We are also missing `ApplyEdgeToEdge` and `LockToOrientationPortrait`.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,805 @@
...
+        }
+    }
+
+    ModalBottomSheet(
+        sheetState = sheetState,
+        onDismissRequest = onDismissRequest,
+        dragHandle = { NostrConnectBottomSheetDragHandle() },
+    ) {
+        BackHandler {
+            val isBudgetPickerVisible =
```
markocic: nitpick: Can we get a different line break?

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,805 @@
...
+                onTabChange = { viewModel.setEvent(NostrConnectContract.UiEvent.TabChanged(it)) },
+                permissionsTabEnabled = state.accounts.isNotEmpty(),
+            )
+
+            NostrConnectPages(
+                state = state,
+                eventPublisher = { viewModel.setEvent(it) },
+            )
+
+            val showDailyBudgetPicker =
```
markocic: This looks like `isBudgetPickerVisible` in `BackHandler`. Can this be a single variable? Preferably wrapped in `remember` and `derivedStateOf`.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,805 @@
...
+            contentAlignment = Alignment.Center,
+        ) {
+            Text(
+                text = stringResource(id = R.string.nostr_connect_no_nsec_accounts_warning),
+                textAlign = TextAlign.Center,
+                style = AppTheme.typography.bodyLarge,
+                color = AppTheme.extraColorScheme.onSurfaceVariantAlt2,
+            )
+        }
+        return
```
markocic: We shouldn't return in Composable functions. This applies to all occurrences of this.

markocic (changes requested):

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -1303,6 +1325,29 @@ private fun NavGraphBuilder.home(
...
+        route = route,
+        arguments = arguments,
+        dialogProperties = DialogProperties(
+            usePlatformDefaultWidth = false,
+        ),
+    ) {
+        val viewModel = hiltViewModel<NostrConnectViewModel>()
+        ApplyEdgeToEdge(isDarkTheme = true)
```
markocic: Why was `isDarkTheme = true` necessary?

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,835 @@
...
+    state: NostrConnectContract.UiState,
+    onDismissRequest: () -> Unit,
+    eventPublisher: (NostrConnectContract.UiEvent) -> Unit,
+    snackbarHostState: SnackbarHostState,
+) {
+    val isBudgetPickerVisible by remember(state.selectedTab, state.showDailyBudgetPicker) {
+        derivedStateOf {
+            state.selectedTab == NostrConnectContract.Tab.PERMISSIONS && state.showDailyBudgetPicker
+        }
+    }
```
markocic, line 153: What does `state.showDailyBudgetPicker` actually indicate? Why do we need this extra check for selected tab?

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,835 @@
...
+
+    BackHandler {
+        if (isBudgetPickerVisible) {
+            eventPublisher(NostrConnectContract.UiEvent.CancelDailyBudget)
+        } else {
+            onDismissRequest()
+        }
+    }
+
+    Box(modifier = Modifier.fillMaxWidth()) {
```
markocic, line 163: Should this be `Scaffold`? Using `SnackbarHost` like this seems weird. Is there a specific reason for `Box` instead of `Scaffold`?

app/src/main/res/values/strings.xml
```diff
@@ -1403,4 +1404,25 @@
 
     <string name="feed_note_render_unknown_audio_title">Unknown title</string>
     <string name="feed_note_render_play_button">Play</string>
+
+    <string name="nostr_connect_unknown_web_app">Unknown Web App</string>
```
markocic: Doesn't have to be a web app. It can be any nostr app that supports nip-46 login.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,835 @@
...
+                    } else {
+                        FontWeight.Normal
+                    },
+                    color = if (permissionsTabEnabled) {
+                        AppTheme.colorScheme.onPrimary
+                    } else {
+                        AppTheme.colorScheme.onSurface.copy(alpha = 0.38f)
+                    },
+                )
+            },
```
markocic: These seem very similar. Could we extract this into another composable to limit copy pasted code and make it easier to change later?

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,835 @@
...
+            val slideIn = if (targetState.ordinal > initialState.ordinal) {
+                primalSlideInHorizontallyFromEnd
+            } else {
+                slideInHorizontally { -it }
+            }
+            val slideOut = if (targetState.ordinal > initialState.ordinal) {
+                primalSlideOutHorizontallyToEnd
+            } else {
+                slideOutHorizontally { it }
+            }
```
markocic: What is different here exactly? Both of them use the same offset. Was it because of the animation spec?

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,835 @@
...
+        supportingContent = account.internetIdentifier?.let {
+            {
+                Text(
+                    text = it.formatNip05Identifier(),
+                    style = AppTheme.typography.bodySmall,
+                    color = AppTheme.extraColorScheme.onSurfaceVariantAlt3,
+                )
+            }
+        },
+        trailingContent = null,
```
markocic: `trailingContent` already has `null` as default.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,835 @@
...
+        )
+    }
+}
+
+@Composable
+private fun NostrConnectPages(
+    state: NostrConnectContract.UiState,
+    eventPublisher: (NostrConnectContract.UiEvent) -> Unit,
+) {
+    AnimatedContent(
```
markocic: I think all of this could be simplified by using `HorizontalPager` instead of `AnimatedContent`. Plus we would get the swipe interaction as well.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectBottomSheet.kt
```diff
@@ -0,0 +1,835 @@
...
+        PrimalFilledButton(
+            modifier = Modifier.weight(1f),
+            height = 50.dp,
+            containerColor = Color.Transparent,
+            contentColor = AppTheme.extraColorScheme.onSurfaceVariantAlt1,
+            border = BorderStroke(width = 1.dp, color = AppTheme.colorScheme.outline),
+            textStyle = AppTheme.typography.bodyLarge.copy(fontWeight = FontWeight.SemiBold, fontSize = 16.sp),
+            onClick = onSecondaryClick,
+        ) {
+            Text(text = secondaryButtonText)
```
markocic, line 805: Should we have some style, colors?

PR #724, opened 2025-10-27
Feature/internal signer keypair

markocic:


app/src/main/kotlin/net/primal/android/auth/repository/AuthRepository.kt
```diff
@@ -22,8 +22,8 @@ class AuthRepository @Inject constructor(
         return userId
     }
 
-    suspend fun loginWithNpub(npub: String, loginType: LoginType): String {
-        val userId = credentialsStore.saveNpub(npub = npub, loginType = loginType)
+    suspend fun loginWithNpub(npub: String, credentialType: CredentialType): String {
+        val userId = credentialsStore.saveNpub(npub = npub, credentialType = credentialType)
```
AleksandarIlic: Do we really need to ask for `credentialType` in repo and in credential store? Or is this because of the Amber?

app/src/main/kotlin/net/primal/android/user/credentials/CredentialsStore.kt
```diff
@@ -32,21 +34,28 @@ class CredentialsStore @Inject constructor(
...
         val (nsec, pubkey) = nostrKey.extractKeyPairFromPrivateKeyOrThrow()
-        addCredential(Credential(nsec = nsec, npub = pubkey, type = LoginType.PrivateKey))
+        addCredential(Credential(nsec = nsec, npub = pubkey, type = CredentialType.PrivateKey))
         return pubkey.bech32ToHexOrThrow()
     }
 
-    suspend fun saveNpub(npub: String, loginType: LoginType): String {
-        addCredential(Credential(nsec = null, npub = npub, type = loginType))
+    suspend fun saveNpub(npub: String, credentialType: CredentialType): String {
+        addCredential(Credential(nsec = null, npub = npub, type = credentialType))
```
AleksandarIlic: If the argument `credentiaType` is because of the Amber, maybe we can have specific functions like we have for `saveNsec` and `getOrCreateInternalSignerCredentials`

PR #726, opened 2025-10-29
Implement NostrConnect Session Active Pill

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/compose/indicator/Indicator.kt
```diff
@@ -0,0 +1,241 @@
...
+    var overlaySize by remember { mutableStateOf(IntSize.Zero) }
+
+    val animatedTopPadding by animateDpAsState(
+        targetValue = topPadding,
+        label = "AnimatedTopPadding",
+    )
+
+    Box(
+        modifier = modifier
+            .offset { IntOffset(offsetX.value.toInt(), offsetY.value.toInt()) }
```
markocic, line 201: Can we get `systemGestureExclusion()` Modifier here as well? I found myself going back while trying to grab the indicator.

app/src/main/kotlin/net/primal/android/core/compose/session/RemoteSessionIndicatorOverlay.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.hilt.navigation.compose.hiltViewModel
+import net.primal.android.R
+import net.primal.android.core.compose.icons.PrimalIcons
+import net.primal.android.core.compose.icons.primaliconpack.NostrConnectSession
+import net.primal.android.core.compose.indicator.IndicatorOverlay
+import net.primal.android.theme.AppTheme
+
+@Composable
+fun RemoteSessionIndicatorOverlay(
+    viewModel: RemoteSessionIndicatorViewModel = hiltViewModel(),
```
AleksandarIlic: We should not hide the view models as default values here. Make it mandatory argument na instantiate it in `PrimalAppNavigation`.

app/src/main/kotlin/net/primal/android/core/compose/session/RemoteSessionIndicatorOverlay.kt
```diff
@@ -0,0 +1,33 @@
...
+    isConnectionUnavailable: Boolean,
```
AleksandarIlic: This name got me confused. I thought it indicate remote session connection. Maybe we could name it `isNetworkAvailable` or `isNetworkUnavailable`. Or something else to avoid confusion with remote session connection(s).

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -486,19 +490,24 @@ fun PrimalAppNavigation(startDestination: String) {
...
     SharedTransitionLayout {
-        ConnectionIndicatorOverlay {
-            PiPManagerProvider {
-                LiveStreamOverlay(
-                    navController = navController,
-                    noteCallbacks = noteCallbacksHandler(navController = navController),
-                ) {
-                    PrimalAppNavigation(
+        ConnectionIndicatorOverlay(viewModel = connectionViewModel) {
+            RemoteSessionIndicatorOverlay(isConnectionUnavailable = !connectionState.hasConnection) {
```
AleksandarIlic: Can we extract `ConnectionIndicatorOverlay` and `RemoteSessionIndicatorOverlay` into an `AppOverlay` composable which instantiate the view models and create this nested structure? This way it's easily extendable in the future for some other overlays.

app/src/main/kotlin/net/primal/android/core/compose/session/RemoteSessionIndicatorViewModel.kt
```diff
@@ -0,0 +1,25 @@
...
+class RemoteSessionIndicatorViewModel @Inject constructor() : ViewModel() {
+
+    private val _state = MutableStateFlow(RemoteSessionIndicatorContract.UiState())
+    val state = _state.asStateFlow()
+    private fun setState(reducer: RemoteSessionIndicatorContract.UiState.() -> RemoteSessionIndicatorContract.UiState) =
+        _state.getAndUpdate(reducer)
+
+    init {
+        viewModelScope.launch {
+            setState { copy(isRemoteSessionActive = true) }
```
AleksandarIlic: You should check if our Android service for remote signer is active or not. If it's active it is `true`, otherwise `false`. @markocic please confirm that this was the decision we made few days ago.

markocic: That's right. Our service will only be active if there is at least one active session.

app/src/main/kotlin/net/primal/android/core/compose/session/RemoteSessionIndicatorViewModel.kt
```diff
@@ -8,18 +8,31 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
+    private val connectionRepository: ConnectionRepository,
+) : ViewModel() {
 
     private val _state = MutableStateFlow(RemoteSessionIndicatorContract.UiState())
     val state = _state.asStateFlow()
     private fun setState(reducer: RemoteSessionIndicatorContract.UiState.() -> RemoteSessionIndicatorContract.UiState) =
         _state.getAndUpdate(reducer)
 
     init {
+        observeRemoteSignerConnections()
```
AleksandarIlic: This is not the correct way. We should observe `PrimalRemoteSignerService` if it's active or not. It's an Android Service we have defined in our app.

PR #734, opened 2025-11-06
Implement `update_push_notification_token_for_nip46` api endpoint

markocic:


app/src/main/kotlin/net/primal/android/user/updater/UserDataUpdater.kt
```diff
@@ -64,7 +64,8 @@ class UserDataUpdater @AssistedInject constructor(
         runCatching { userRepository.fetchAndUpdateUserAccount(userId = userId) }
         runCatching { bookmarksRepository.fetchAndPersistBookmarks(userId = userId) }
         runCatching { primalWalletAccountRepository.fetchWalletAccountInfo(userId = userId) }
-        runCatching { pushNotificationsTokenUpdater.updateTokenForAllUsers() }
+        pushNotificationsTokenUpdater.updateTokenForAllUsers()
+        pushNotificationsTokenUpdater.updateTokenForNip46()
```
AleksandarIlic: Why is this not in runCatching? Even though it's not crashing let's wrap it up to avoid confusion when looking at the code.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectViewModel.kt
```diff
@@ -178,6 +180,7 @@ class NostrConnectViewModel @Inject constructor(
                 userPubKey = selectedAccount.pubkey,
                 connectionUrl = connectionUrl,
             ).onSuccess {
+                tokenUpdater.updateTokenForNip46()
```
AleksandarIlic: If this breaks, the execution should continue. Wrap it in runCatching. If underlying code changes we must be sure that this is not broken by accident.

AleksandarIlic (changes requested):
Please use ForRemoteSigner instead of forNip46 everywhere.

app/src/main/kotlin/net/primal/android/settings/connected/details/ConnectedAppDetailsViewModel.kt
```diff
@@ -113,6 +115,7 @@ class ConnectedAppDetailsViewModel @Inject constructor(
             }.onSuccess {
                 setState { copy(confirmingDeletion = false) }
                 setEffect(SideEffect.ConnectionDeleted)
+                runCatching { tokenUpdater.updateTokenForRemoteSigner() }
```
AleksandarIlic: This could be cancelled before it finishes. Place it in the `runCatching` block above, after `deleteConnection`

PR #749, opened 2025-11-19
Refactor PermissionsBottomSheet to use core UI utilities

mehmedalijaK:


app/src/main/kotlin/net/primal/android/nostrconnect/permissions/PermissionsBottomSheet.kt
```diff
@@ -286,14 +281,17 @@ fun AppRequestListItem(
...
 
-    if (!isLast) {
+    if (showDivider) {
         PrimalDivider()
     }
 }
 
 @Composable
 private fun SessionEventMetadata(requestTypeId: String, requestedAt: Long) {
-    val formattedDateTime = rememberFormattedDateTime(timestamp = requestedAt)
```
markocic, line 296: Is this unused now? I think it is. If so, delete it.

markocic (changes requested):

PR #750, opened 2025-11-20
Implement event details view for pending requests in permissions sheet

mehmedalijaK:


app/src/main/kotlin/net/primal/android/nostrconnect/permissions/PermissionsViewModel.kt
```diff
@@ -60,10 +63,42 @@ class PermissionsViewModel @Inject constructor(
...
                 }
             }
         }
 
+    private fun handleOpenEventDetails(eventId: String) {
+        val sessionEvent = state.value.sessionEvents.find { it.eventId == eventId }
+        var nostrEvent: NostrEvent? = null
+
+        if (sessionEvent is SessionEvent.SignEvent) {
+            val rawJson = sessionEvent.signedNostrEventJson?.ifEmpty { null }
```
markocic: Let's lose `signedNostrEventJson`.

data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/mappers/SessionEventMappers.kt
```diff
@@ -105,6 +108,24 @@ fun SessionEventData.asDomain(): SessionEvent? {
 
         SignerMethodType.SignEvent -> {
             val eventKind = this.eventKind?.decrypted ?: return null
+            val requestMethod = requestPayload?.decodeFromJsonStringOrNull<RemoteSignerMethod>()
+
+            val unsignedEventJson = if (requestMethod is RemoteSignerMethod.SignEvent) {
+                val unsignedEvent = requestMethod.unsignedEvent.withPubKey(this.signerPubKey.decrypted)
+                val dummyEvent = NostrEvent(
```
markocic: We have `UnsignedNostrEvent` that doesn't require `id` and `sig`. We should use that one as it better reflects the stored data.

app/src/main/kotlin/net/primal/android/nostrconnect/ui/NostrEventDetails.kt
```diff
@@ -64,9 +64,10 @@ private sealed interface EventDetailRow {
 @Composable
 fun NostrEventDetails(
     event: NostrEvent,
```
markocic: Since we are now showing `UnsignedNostrEvent` and `NostrEvent` we can change this parameter to take `List<EventDetailRow>` and we can build what we want to be shown when calling it.

app/src/main/kotlin/net/primal/android/core/compose/nostr/NostrEventDetails.kt
```diff
@@ -119,8 +120,10 @@ fun NostrEventDetails(
...
         }
 
-        if (rawJson != null) {
-            item(key = "CopyRawJsonButton", contentType = "Button") {
+        item(key = "Footer", contentType = "Button") {
+            if (footerContent != null) {
+                footerContent()
+            } else if (rawJson != null) {
                 val rawJsonLabel = stringResource(id = R.string.settings_event_details_raw_json_label)
                 PrimalFilledButton(
```
markocic: What do you think about forwarding this logic through `footerContent` where you wanted copy button to appear. This behaviour isn't immediately obvious reading function arguments.

app/src/main/kotlin/net/primal/android/nostrconnect/permissions/PermissionsViewModel.kt
```diff
@@ -60,10 +63,42 @@ class PermissionsViewModel @Inject constructor(
...
+                ?: sessionEvent.unsignedNostrEventJson
+
+            nostrEvent = rawJson.let {
+                runCatching {
+                    NostrJsonEncodeDefaults.decodeFromString<NostrEvent>(it)
```
markocic: This should be `UnsignedNostrEvent`.

app/src/main/kotlin/net/primal/android/nostrconnect/permissions/PermissionsContract.kt
```diff
@@ -10,6 +11,8 @@ interface PermissionsContract {
         val requestQueue: List<Pair<ActiveSessionUi, List<SessionEvent>>> = emptyList(),
         val activeSessions: Map<String, ActiveSessionUi> = emptyMap(),
         val responding: Boolean = false,
+        val eventDetails: SessionEvent? = null,
```
markocic: Do we need this in state? All we are doing is checking if it is `null`. Can't we use `eventDetailsNostrEvent` for it?

markocic (changes requested):

PR #765, opened 2025-11-28
Bugfix/scope leaks

markocic:


data/account/remote/src/commonMain/kotlin/net/primal/data/account/remote/client/RemoteSignerClient.kt
```diff
@@ -93,11 +94,10 @@ class RemoteSignerClient(
         }
     }
 
-    fun close() =
-        scope.launch {
-            nostrSocketClient.close()
-            listenerJob?.cancel()
-        }
+    suspend fun close() {
```
AleksandarIlic: This is going to make the instance unusable later since we are canceling the scope. Maybe we should consider naming it `destroy()`? If we need additional reconnects, maybe we should add `disconnect()` additionally.

core/networking-primal/src/commonMain/kotlin/net/primal/core/networking/primal/ProxyPrimalApiClient.kt
```diff
@@ -127,6 +131,7 @@ internal class ProxyPrimalApiClient(
 
     override suspend fun closeSubscription(subscriptionId: String): Boolean {
         clientInitialized.await()
+        observeApiUrlJob?.cancel()?.also { observeApiUrlJob = null }
```
AleksandarIlic: I think we should not do this. This will break this proxy client if just one subscription is closed. We can have here multiple subscriptions and also specific queries without subscriptions.

data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/manager/NostrRelayManager.kt
```diff
@@ -124,8 +126,11 @@ internal class NostrRelayManager(
 
     private fun observeClientMethods(relay: String, client: RemoteSignerClient) {
         clients[relay] = client
-        clientJobs[relay] = scope.launch {
-            scope.launch {
+
+        var methodsJob: Job? = null
+        var errorsJob: Job? = null
+        val parentJob = scope.launch {
```
AleksandarIlic: What is the goal here?

If parent job should fail if any children job fails, then you should use just `launch {}` instead of `scope.launch {}`

If you want children to fail independently, the you can use
```
parentScope.launch {
    supervisorScope {
        launch { flow1.collect { } }
        launch { flow2.collect { } }
    }
}
```

With this approach you get to keep only one Job instead of list of Jobs since you are interested in cancelling everything and not particular child jobs.

PR #770, opened 2025-12-01
Implement background avatar pre-caching with isolated storage

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/di/ImagesModule.kt
```diff
@@ -0,0 +1,24 @@
...
+import net.primal.core.caching.MediaCacher
+
+@Module
+@InstallIn(SingletonComponent::class)
+object ImagesModule {
+
+    @Provides
+    @Singleton
+    fun provideMediaCacher(@ApplicationContext context: Context): MediaCacher {
+        val avatarImageLoader = AvatarCoilImageLoader.provideImageLoader(context)
```
markocic, line 22: We have two image loaders we use for avatars. `provideImageLoader` and `provideNoGifsImageLoader`. Should we distinguish between the two here?

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+
+private const val PRE_CACHE_DEBOUNCE_DELAY = 500L
+
+@Singleton
+class CoilMediaCacher @Inject constructor(
+    @ApplicationContext private val context: Context,
+    private val imageLoader: ImageLoader,
+) : MediaCacher {
+
+    private val scope = CoroutineScope(Dispatchers.IO + SupervisorJob())
```
markocic: We should use `dispatchers.io()` instead of `Dispatchers.IO`.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+
+        preCacheJob = scope.launch {
+            delay(PRE_CACHE_DEBOUNCE_DELAY)
+
+            val uniqueUrls = urls
+                .filter { it.isNotBlank() }
+                .distinct()
+
+            for (url in uniqueUrls) {
+                if (!isActive) break
```
markocic: Why is this necessary?

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+                val request = ImageRequest.Builder(context)
+                    .data(url)
+                    .memoryCachePolicy(CachePolicy.DISABLED)
+                    .diskCachePolicy(CachePolicy.ENABLED)
+                    .networkCachePolicy(CachePolicy.ENABLED)
+                    .build()
+
+                imageLoader.enqueue(request)
+
+                yield()
```
markocic: What effect does `yield` have here?

core/caching/build.gradle.kts
```diff
@@ -0,0 +1,69 @@
+import org.jetbrains.kotlin.gradle.plugin.mpp.apple.XCFramework
+
+plugins {
+    alias(libs.plugins.kotlin.multiplatform)
+    alias(libs.plugins.android.kotlin.multiplatform.library)
+}
+
+private val xcfName = "PrimalCaching"
```
markocic: `PrimalCoreCaching`.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.yield
+import net.primal.core.caching.MediaCacher
+
+private const val PRE_CACHE_DEBOUNCE_DELAY = 500L
+
+@Singleton
+class CoilMediaCacher @Inject constructor(
+    @ApplicationContext private val context: Context,
```
AleksandarIlic: Please avoid storing `Context` as a field in the classes. Use it as constructor field only.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.yield
+import net.primal.core.caching.MediaCacher
+
+private const val PRE_CACHE_DEBOUNCE_DELAY = 500L
```
AleksandarIlic: If this is part of the `CoilMediaCacher` then it should be in a `private companion object` of that class.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/feed/paging/NoteFeedRemoteMediator.kt
```diff
@@ -160,6 +163,10 @@ internal class NoteFeedRemoteMediator(
         ) {
             val response = withContext(dispatcherProvider.io()) { feedApi.getFeedBySpec(body = requestBody) }
             response.paging ?: throw NetworkException("PagingEvent not found.")
+
+            val avatarUrls = response.metadata.mapAsAvatarUrls(cdnResources = response.cdnResources)
+            mediaCacher.preCacheUserAvatars(avatarUrls)
```
AleksandarIlic: This block is repeating. Can you create it as an extension function and call the function?

AleksandarIlic: Actually, you need to define a caching util somewhere in this module as a public `fun MediaCacher.cacheAvatarUrls(metadata: List<NostrEvent>, cdnResources: List<PrimalEvent>)` and on every place just call this. The mapping logic from `MetadataEvents` should be also moved there.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+                .distinct()
+
+            for (url in uniqueUrls) {
+                if (!isActive) break
+
+                val request = ImageRequest.Builder(context)
+                    .data(url)
+                    .memoryCachePolicy(CachePolicy.DISABLED)
+                    .diskCachePolicy(CachePolicy.ENABLED)
+                    .networkCachePolicy(CachePolicy.ENABLED)
```
AleksandarIlic: We might not need network cache since we are already caching it to the disk. No need to cache it twice.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
AleksandarIlic: `enqueue` schedules the download in the background. I guess `yield()` was to give some room for other coroutines, but unfortunately this has no effect because the actual download is happening in different place. So `yield()` is unnecessary here.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+) : MediaCacher {
+
+    private val scope = CoroutineScope(Dispatchers.IO + SupervisorJob())
+
+    private var preCacheJob: Job? = null
+
+    override fun preCacheUserAvatars(urls: List<String>) {
+        if (urls.isEmpty()) return
+
+        preCacheJob?.cancel()
```
AleksandarIlic: If multiple repositories call `preCacheUserAvatars` at the same time only the last one will survive, which should not be the case. I suggest you don't store the `Job` and instead limit the life of started coroutine to 20 seconds (const in companion object) using `withTimeout() {}`.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -0,0 +1,60 @@
...
+
+        preCacheJob = scope.launch {
+            delay(PRE_CACHE_DEBOUNCE_DELAY)
```
AleksandarIlic: Why are you delaying the start?

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
AleksandarIlic: `enqueue` does not download the image, so this is also not necessary. We have two options:
1) to delegate parallelism to Coil using `enqueue`
2) or to manual control it using `execute`

app/src/main/kotlin/net/primal/android/core/di/ImagesModule.kt
mehmedalijaK: Good point. However, since we only store raw data when caching to disk (Memory Cache is disabled), this approach is fine. We aren't decoding images in the Cacher, and since both loaders share the same disk cache folder, the file will be available for both.

The UI layer will then decide which loader to use for displaying the image based on user settings, but both will pull from the same file we just cached.

app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
mehmedalijaK: I think we actually need `.networkCachePolicy(CachePolicy.ENABLED)`.
 
If I understand the docs correctly, disabling this would prevent network requests if the image isn't already on disk. Since the purpose of this Cacher is to download missing images, we must keep the network policy enabled.

app/src/main/kotlin/net/primal/android/core/di/ImagesModule.kt
AleksandarIlic: This sounds reasonable. I'm resolving this comment.

core/caching/build.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+private val xcfName = "PrimalCoreCaching"
+
+kotlin {
+    // Android target
+    androidLibrary {
+        namespace = "net.primal"
+        compileSdk = 35
+        minSdk = 26
```
AleksandarIlic: Please check latest `main` for this.

core/caching/src/commonMain/kotlin/net/primal/core/caching/MediaCacher.kt
```diff
@@ -0,0 +1,5 @@
+package net.primal.core.caching
+
+interface MediaCacher {
```
AleksandarIlic, line 3: Wherever we use this class in repositories, can we make it optional? Like `mediaCache: MediaCacher? = null`. We can there use it as `mediaCacher?.` do what we want. Technically, this caching stuff is not mandatory, so we should not force it.

AleksandarIlic (changes requested):

PR #773, opened 2025-12-01
Enhance Active Sessions selection logic and settings navigation

mehmedalijaK:


app/src/main/kotlin/net/primal/android/nostrconnect/list/ActiveSessionsViewModel.kt
```diff
@@ -75,12 +81,24 @@ class ActiveSessionsViewModel @Inject constructor(
...
+        val selectedSessionIds = state.value.selectedSessions
+        if (selectedSessionIds.size == 1) {
+            val sessionId = selectedSessionIds.first()
+            val session = state.value.sessions.find { it.sessionId == sessionId }
+            setEffect(SideEffect.NavigateToConnectedApps(connectionId = session?.connectionId))
+        } else {
+            setEffect(SideEffect.NavigateToConnectedApps(connectionId = null))
+        }
+    }
+
```
markocic: Screen should handle this. Not ViewModel.

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
```diff
@@ -1382,7 +1382,14 @@ private fun NavGraphBuilder.activeSessions(navController: NavController) {
...
             onDismissRequest = { navController.popBackStack() },
-            onSettingsClick = { navController.navigateToConnectedApps() },
+            onSettingsClick = { connectionId ->
+                navController.popBackStack()
+                if (connectionId != null) {
+                    navController.navigateToConnectedAppDetails(connectionId = connectionId)
+                } else {
+                    navController.navigateToConnectedApps()
+                }
+            },
```
markocic, line 1392: Screen should get separate callbacks and distinguish between the two. This requires the caller to know about the business logic that happens on `onSettingsClick`.

markocic (changes requested):

app/src/main/kotlin/net/primal/android/navigation/PrimalAppNavigation.kt
AleksandarIlic: Actually, let's keep this logic like this.

app/src/main/kotlin/net/primal/android/nostrconnect/list/ActiveSessionsViewModel.kt
AleksandarIlic: I agree. We can handle this on UI since everything is in the UI state. No need to go through ViewModel since not gets validated or approved.

AleksandarIlic (changes requested):
- Remove side effects.
- Move handling settings click to UI.
- Keep settings callback in app navigation as they are right now.

PR #775, opened 2025-12-01
Implement reset permissions to default for connected apps

mehmedalijaK:


app/src/main/kotlin/net/primal/android/settings/connected/permissions/AppPermissionsScreen.kt
```diff
@@ -155,6 +120,75 @@ fun AppPermissionsScreen(
             )
         }
     }
+
+    if (state.confirmingReset) {
```
markocic: Visibility should be controlled on the UI. No need for visibility changes to go through `ViewModel`. We are not doing anything there.

data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/repository/PermissionsRepositoryImpl.kt
```diff
@@ -68,6 +69,35 @@ class PermissionsRepositoryImpl(
...
+                    AppPermissionData(
+                        permissionId = permissionId,
+                        connectionId = connectionId,
+                        action = action,
+                    )
+                }
+
+                database.permissions().upsertAll(resetPermissions)
+            }
+        }
```
markocic, line 91: I don't understand why we need this complex logic. Why didn't we just set current permissions to medium trust ones? That should be enough.

AleksandarIlic: I also don't get the logic. The expected behaviour should be:
If user confirms the reset, then all local permissions should be overwritten by default permissions (which are all `allow`).

app/src/main/kotlin/net/primal/android/settings/connected/permissions/AppPermissionsScreen.kt
AleksandarIlic: I agree. Only `ResetPermissions` event should be added to the contract which gets called when users confirms the reset on dialog.

markocic (approved):

AleksandarIlic (approved):

PR #779, opened 2025-12-03
Implement session inactivity timeout

markocic:


data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/service/RemoteSignerServiceImpl.kt
```diff
@@ -62,7 +88,10 @@ class RemoteSignerServiceImpl internal constructor(
                             valueTransform = { it.second },
                         )
 
-                    sessions.forEach { setActiveRelays(it) }
+                    sessions.forEach {
+                        setActiveRelays(it)
+                        sessionActivityMap.getOrPut(it.sessionId) { Clock.System.now() }
```
AleksandarIlic, line 93: Why did you choose custom activity logging instead of using timestamps from a session?

markocic: Just a scalability issue. This way we are only having a few timestamps in memory at all times but session events will scale the longer the user uses our signer. It can potentially scale to a large number of rows in database and at that point this will be a more efficient way. Worst case scenario - they are the same, best case scenario - in memory is more optimal.

It's worth noting that this will probably affect a small number of users and I'm not sure to what extent. It could be noticeable, or insignificant. I am okay with changing it if this isn't a trade-off we want.

AleksandarIlic: It is fine as it is, I was just curios. Are you sure `Session` is updated on every request? We want to auto-close session if there are no new REQUESTS within the defined inactivity period. For example user is active for 25 minutes and then stops using the web app. If there are no new requests for 15 minutes the session should be stopped - this is exactly 40 minutes after the session was started.

PR #783, opened 2025-12-04
Fix race condition when initializing new connection

markocic:


data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/manager/NostrRelayManager.kt
```diff
@@ -50,10 +51,13 @@ internal class NostrRelayManager(
     private val _relayEvents = MutableSharedFlow<RelayEvent>(extraBufferCapacity = 64)
     val relayEvents = _relayEvents.asSharedFlow()
 
+    private val firstConnect = CompletableDeferred<Unit>()
```
AleksandarIlic: This should not be in NostrRelayManager. Looks like this needs to be handled in the remote service.

PR #784, opened 2025-12-04
Overwrite existing connection transactionally on save

mehmedalijaK:


data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/repository/ConnectionRepositoryImpl.kt
```diff
@@ -73,6 +73,16 @@ class ConnectionRepositoryImpl(
...
+                val existing = database.connections().getConnectionByClientPubKey(
+                    clientPubKey = connection.clientPubKey.asEncryptable(),
+                )
+                if (existing != null && existing.data.connectionId != connection.connectionId) {
+                    val connectionId = existing.data.connectionId
+                    database.sessions().deleteSessionsByConnectionId(connectionId)
+                    database.permissions().deletePermissionsByConnectionId(connectionId)
+                    database.connections().deleteConnection(connectionId)
+                }
+
```
AleksandarIlic: - Try inserting first. 
- If it fails then use existing function to delete connection and then insert again. Extract in private fun if necessary.
- If it succeeds then the work is done.

AleksandarIlic: And add deleting session events in `delete` and the end.

data/account/local/src/commonMain/kotlin/net/primal/data/account/local/dao/SessionEventDataDao.kt
```diff
@@ -39,4 +39,11 @@ interface SessionEventDataDao {
 
     @Query("SELECT * FROM SessionEventData WHERE eventId = :eventId")
     suspend fun getSessionEvent(eventId: String): SessionEventData?
+
+    @Query(
+        """
+        DELETE FROM SessionEventData 
+        WHERE sessionId IN (SELECT sessionId FROM AppSessionData WHERE connectionId = :connectionId)""",
```
markocic: We can just delete by `clientPubKey`.

data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/repository/ConnectionRepositoryImpl.kt
```diff
@@ -72,26 +71,21 @@ class ConnectionRepositoryImpl(
...
-                )
-
-                database.permissions().upsertAll(data = connection.permissions.map { it.asPO() })
+            try {
+                database.withTransaction {
+                    saveConnectionInternal(secret, connection)
+                }
+            } catch (_: SQLiteException) {
+                database.withTransaction {
+                    val existing = database.connections().getConnectionByClientPubKey(
```
markocic: Why do you have another check if this exists? If we are in `catch` block it means we have this connection and were unable to insert.

data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/repository/ConnectionRepositoryImpl.kt
```diff
@@ -168,4 +162,30 @@ class ConnectionRepositoryImpl(
                 }
             }
         }
+
+    private suspend fun saveConnectionInternal(secret: String, connection: AppConnection) {
```
markocic: Not sure about these names. `private` already indicates that this is an internal function.

PR #789, opened 2025-12-08
Implement approve & reject for all events

mehmedalijaK:


app/src/main/kotlin/net/primal/android/nostrconnect/ui/EventDetailsUiHelper.kt
```diff
@@ -0,0 +1,237 @@
...
+import net.primal.android.R
+import net.primal.domain.account.model.RequestState
+import net.primal.domain.account.model.SessionEvent
+import net.primal.domain.nostr.NostrEvent
+import net.primal.domain.nostr.NostrUnsignedEvent
+
+val NostrConnectSuccessColor = Color(0xFF40FF00)
+val NostrConnectRejectedColor = Color(0xFFFFA02F)
+
+object EventDetailsUiHelper {
```
AleksandarIlic: If this is for Compose, you don't need an object for all this ui helper functions, right?

app/src/main/kotlin/net/primal/android/nostrconnect/ui/EventDetailsUiHelper.kt
```diff
@@ -0,0 +1,237 @@
...
+
+import android.content.Context
+import androidx.compose.ui.graphics.Color
+import net.primal.android.R
+import net.primal.domain.account.model.RequestState
+import net.primal.domain.account.model.SessionEvent
+import net.primal.domain.nostr.NostrEvent
+import net.primal.domain.nostr.NostrUnsignedEvent
+
+val NostrConnectSuccessColor = Color(0xFF40FF00)
```
AleksandarIlic: It looks really bad on light theme. We can't hard-code it actually. You should use from theme. Looks like `AppTheme.extraColorScheme.successBright` might be a good fit.

app/src/main/kotlin/net/primal/android/nostrconnect/ui/EventDetailsUiHelper.kt
```diff
@@ -0,0 +1,237 @@
...
+val NostrConnectRejectedColor = Color(0xFFFFA02F)
```
AleksandarIlic, line 13: We can keep this hard-coded. It's actually `sunriseZapped` from theme, but let's keep hard-coded to avoid confusion.

AleksandarIlic (changes requested):
This looks great. Just minor color tweaks. You might have to change the logic there since now you have to read from theme, but should not be a big deal.

AleksandarIlic (approved):

PR #790, opened 2025-12-09
Fix PiP buffering behavior

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/home/HomeFeedScreen.kt
```diff
@@ -78,6 +81,12 @@ fun HomeFeedScreen(
 ) {
     val streamState = LocalStreamState.current
     val uiState = viewModel.state.collectAsState()
+    val pipManager = LocalPiPManager.current
+    val activity = LocalActivity.current
+
+    BackHandler(enabled = pipManager.shouldEnterPiPMode) {
+        activity?.enterPictureInPictureMode(pipManager.buildPipParams())
+    }
```
AleksandarIlic: Remove this. We don't want PIP on back.

app/src/main/kotlin/net/primal/android/stream/LiveStreamOverlay.kt
```diff
@@ -122,7 +122,9 @@ private fun LiveStreamOverlay(
                         viewModel.setEvent(UiEvent.OnPlayerStateUpdate(totalDuration = duration))
                     }
                 }
-
+                Player.STATE_BUFFERING -> {
+                    pipManager.shouldEnterPiPMode = true
+                }
```
AleksandarIlic, line 127: We can keep this improvement.

PR #801, opened 2025-12-12
Implement media preloading for smoother scrolling

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/images/CoilMediaCacher.kt
```diff
@@ -41,7 +42,30 @@ class CoilMediaCacher @Inject constructor(
...
+                for (url in uniqueUrls) {
+                    val request = ImageRequest.Builder(context)
+                        .data(url)
+                        .memoryCachePolicy(CachePolicy.ENABLED)
+                        .diskCachePolicy(CachePolicy.ENABLED)
+                        .networkCachePolicy(CachePolicy.ENABLED)
+                        .build()
+
+                    feedImageLoader.enqueue(request)
                 }
```
AleksandarIlic, line 66: You could extract this into a private function which accepts `ImageLoader` and `ImageRequest.Builder`  and reuse it in both places.

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -173,17 +181,106 @@ private fun NoteFeedList(
...
+            listState = listState,
+        )
+
+        TopVisibleNoteTracker(
+            listState = listState,
+            pagingItems = pagingItems,
+            eventPublisher = eventPublisher,
+        )
+
+        Box {
```
AleksandarIlic: Do we really need `Box` within `BoxWithConstraints`? Maybe just the box with constraints is enough?

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -173,17 +181,106 @@ private fun NoteFeedList(
...
+        val feedWidthPx = with(LocalDensity.current) { constraints.maxWidth }
+
+        FeedMediaPreloader(
+            pagingItems = pagingItems,
+            listState = listState,
+            mediaCacher = mediaCacher,
+            feedWidthPx = feedWidthPx,
+        )
+
+        ScrollToTopHandler(
```
AleksandarIlic: `FeedMediaPreloader` and `ScrollToTopHandler` could go outside of the Box?

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -484,3 +532,85 @@ private fun NewPillStringIndicator(
...
+    }
+}
+
+private const val MAX_VISIBLE_MEDIA_ATTACHMENTS = 4
+private const val MEDIA_ATTACHMENT_COUNT_ONE = 1
+private const val MEDIA_ATTACHMENT_COUNT_TWO = 2
+private const val MEDIA_ATTACHMENT_COUNT_THREE = 3
+
+private fun FeedPostUi.extractMediaUrls(feedWidthPx: Int): List<String> {
+    val mediaAttachments = this.uris.filter { it.isMediaUri() }
```
AleksandarIlic: We should maybe include link previews and take only first two if there are more.

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -484,3 +532,85 @@ private fun NewPillStringIndicator(
...
+            MAX_VISIBLE_MEDIA_ATTACHMENTS,
+        )
+    } else {
+        mediaAttachments
+    }
+
+    return visibleAttachments.mapIndexedNotNull { index, uri ->
+        when (uri.type) {
+            EventUriType.Image -> {
+                val targetWidth = when (totalCount) {
```
AleksandarIlic: I think we don't need this logic for variants. In `NoteMediaAttachmentsHorizontalPager` we use `val imageSizeDp = findImageSize(eventUri = mediaEventUris.first())` for image size.

app/src/main/kotlin/net/primal/android/notes/feed/list/NoteFeedList.kt
```diff
@@ -484,3 +532,85 @@ private fun NewPillStringIndicator(
...
+            }
+    }
+}
+
+private const val MAX_VISIBLE_MEDIA_ATTACHMENTS = 4
+private const val MEDIA_ATTACHMENT_COUNT_ONE = 1
+private const val MEDIA_ATTACHMENT_COUNT_TWO = 2
+private const val MEDIA_ATTACHMENT_COUNT_THREE = 3
+
+private fun FeedPostUi.extractMediaUrls(feedWidthPx: Int): List<String> {
```
AleksandarIlic, line 591: The main concern I have with logic of this function is that we have copies of business logic. If we change the business logic on max visible media attachments (gallery of 4) then we have to remember to come here and update it. The same applies for how variant is resolved - if it changes over there, we have to follow-up here. Could we somehow reuse the same business logic and/or constants? So when it changes somewhere there, to reflect immediately here as well.

AleksandarIlic (approved):

PR #802, opened 2025-12-18
Extract SignerConnectBottomSheet from NostrConnect flow

mehmedalijaK:


app/src/main/kotlin/net/primal/android/nostrconnect/signer/SignerConnectBottomSheet.kt
```diff
@@ -0,0 +1,705 @@
+package net.primal.android.nostrconnect.signer
```
AleksandarIlic: This needs and it's dependencies should be in `core` since it's going to be reused between 2 different features.

app/src/main/kotlin/net/primal/android/nostrconnect/signer/SignerConnectBottomSheet.kt
```diff
@@ -0,0 +1,705 @@
...
+import net.primal.android.navigation.primalSlideInHorizontallyFromEnd
+import net.primal.android.navigation.primalSlideOutHorizontallyToEnd
+import net.primal.android.theme.AppTheme
+import net.primal.domain.account.model.TrustLevel
+import net.primal.domain.links.CdnImage
+
+@Composable
+fun SignerConnectBottomSheet(
+    state: SignerConnectContract.UiState,
+    eventPublisher: (SignerConnectContract.UiEvent) -> Unit,
```
AleksandarIlic: I'm not sure if we should reuse the contract and event publisher. I don't like the idea that viewModels are going to be dependant on this contract.

AleksandarIlic: Yeah, let's decouple this. We should:
- extract state as ui model for this bottom sheet
- replace event publisher with callbacks (the event publisher is a concept we use with view model) - if necessary you can put them in a data class (if there are many).

AleksandarIlic (changes requested):
This needs more decoupling.

app/src/main/kotlin/net/primal/android/core/compose/signer/model/SignerConnectCallbacks.kt
```diff
@@ -0,0 +1,16 @@
+package net.primal.android.core.compose.signer.model
+
+import net.primal.domain.account.model.TrustLevel
+
+data class SignerConnectCallbacks(
```
AleksandarIlic: We only need `onConnect` and `onCancel`.

app/src/main/kotlin/net/primal/android/core/compose/signer/model/SignerConnectUiState.kt
```diff
@@ -0,0 +1,34 @@
...
+    // val dailyBudget: Long? = null,
+    // val showDailyBudgetPicker: Boolean = false,
+    // val selectedDailyBudget: Long? = null,
+    // val budgetToUsdMap: Map<Long, BigDecimal?> = emptyMap(),
+)
+
+enum class SignerConnectTab {
+    Login,
+    Permissions,
+}
```
AleksandarIlic: This could be internal in the composable file.

app/src/main/kotlin/net/primal/android/core/compose/signer/model/SignerConnectUiState.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import net.primal.android.core.errors.UiError
+import net.primal.android.drawer.multiaccount.model.UserAccountUi
+import net.primal.domain.account.model.TrustLevel
+
+data class SignerConnectUiState(
+    val appName: String?,
+    val appDescription: String?,
+    val appImageUrl: String?,
+    val connectionUrl: String?,
```
AleksandarIlic: Remove this.

app/src/main/kotlin/net/primal/android/core/compose/signer/model/SignerConnectUiState.kt
```diff
@@ -0,0 +1,34 @@
...
+    val accounts: List<UserAccountUi> = emptyList(),
+    val selectedTab: SignerConnectTab = SignerConnectTab.Login,
+    val selectedAccount: UserAccountUi? = null,
+    val trustLevel: TrustLevel = TrustLevel.Medium,
+    val connecting: Boolean = false,
+    val error: UiError? = null,
```
AleksandarIlic: Error should also not be part of the UI. Screens which use this component should handle errors.

app/src/main/kotlin/net/primal/android/core/compose/signer/model/SignerConnectUiState.kt
```diff
@@ -0,0 +1,34 @@
...
+import net.primal.android.drawer.multiaccount.model.UserAccountUi
+import net.primal.domain.account.model.TrustLevel
+
+data class SignerConnectUiState(
+    val appName: String?,
+    val appDescription: String?,
+    val appImageUrl: String?,
+    val connectionUrl: String?,
+    val accounts: List<UserAccountUi> = emptyList(),
+    val selectedTab: SignerConnectTab = SignerConnectTab.Login,
```
AleksandarIlic: Selected tab should be internal in the component.

app/src/main/kotlin/net/primal/android/core/compose/signer/model/SignerConnectUiState.kt
```diff
@@ -0,0 +1,34 @@
...
+    val selectedAccount: UserAccountUi? = null,
+    val trustLevel: TrustLevel = TrustLevel.Medium,
```
AleksandarIlic: This should be internal so that the caller should not care about trust level. It's going to get the info from onConnect callback.

app/src/main/kotlin/net/primal/android/core/compose/signer/model/SignerConnectUiState.kt
```diff
@@ -0,0 +1,34 @@
...
+import net.primal.domain.account.model.TrustLevel
+
+data class SignerConnectUiState(
+    val appName: String?,
+    val appDescription: String?,
+    val appImageUrl: String?,
+    val connectionUrl: String?,
+    val accounts: List<UserAccountUi> = emptyList(),
+    val selectedTab: SignerConnectTab = SignerConnectTab.Login,
+    val selectedAccount: UserAccountUi? = null,
```
AleksandarIlic: Let's just remove this. Keep this internal. Preselect first account and then keep the state internal.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectContract.kt
```diff
@@ -1,53 +1,33 @@
...
-    enum class Tab {
-        LOGIN,
-        PERMISSIONS,
-    }
-
     sealed class UiEvent {
-        data class ChangeTab(val tab: Tab) : UiEvent()
+        data class ChangeTab(val tab: SignerConnectTab) : UiEvent()
         data class SelectAccount(val pubkey: String) : UiEvent()
         data class SelectTrustLevel(val level: TrustLevel) : UiEvent()
```
AleksandarIlic: This is going to be part of the internal component.

AleksandarIlic (approved):
This looks great now! Good work!
@markocic Please double check in the morning.

app/src/main/kotlin/net/primal/android/nostrconnect/NostrConnectContract.kt
```diff
@@ -7,47 +7,30 @@ import net.primal.domain.account.model.TrustLevel
...
-        data class ChangeTab(val tab: Tab) : UiEvent()
-        data class SelectAccount(val pubkey: String) : UiEvent()
-        data class SelectTrustLevel(val level: TrustLevel) : UiEvent()
-
-        // data object ClickDailyBudget : UiEvent()
-        // data class ChangeDailyBudget(val budget: Long?) : UiEvent()
-        // data object ApplyDailyBudget : UiEvent()
-        // data object CancelDailyBudget : UiEvent()
-        data object ClickConnect : UiEvent()
+        data class ClickConnect(val account: UserAccountUi, val trustLevel: TrustLevel) : UiEvent()
```
AleksandarIlic, line 23: Instead of account, we actually just need `userId: String`.

PR #803, opened 2025-12-19
Fix NostrRelayManager concurrency and autoStart SQL query error

mehmedalijaK:


data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/manager/NostrRelayManager.kt
```diff
@@ -159,10 +166,14 @@ internal class NostrRelayManager(
...
     }
 
     private suspend fun removeClient(relay: String) {
-        clients.remove(relay)?.destroy()
-        clientJobs.remove(relay)?.cancel()
+        val oldClients = clients.fetchAndUpdate { it - relay }
+        oldClients[relay]?.destroy()
+
+        val oldJobs = clientJobs.fetchAndUpdate { it - relay }
+        oldJobs[relay]?.cancel()
```
AleksandarIlic: Let's use utils to avoid fetchAndUpdate.

AleksandarIlic (changes requested):

PR #813, opened 2025-12-23
Fix duplicate session creation race condition

mehmedalijaK:


AleksandarIlic (approved):
The issue is that we start processing incoming methods in parallel while session is still not marked as active. As a result session `autoStart` features kicks in and start creating new sessions. This changes are ensuring that only one open (not-ended) session can exist at one time and therefore prevents opening new sessions.

`findActiveSessionId` in remote signer client is still looking for active session id when storing request. In case of longer delays, we can end up in a situation where processed event would not be logged. We have to look into this additionally how to solve.

PR #820, opened 2025-12-26
Implement enable notifications to keep running remote signer in the background on first connection

mehmedalijaK:


AleksandarIlic (approved):
LGTM!

PR #835, opened 2026-01-05
Implement `name` field in `LocalApp`

markocic:


data/account/local/src/commonMain/kotlin/net/primal/data/account/local/db/AccountDatabase.kt
```diff
@@ -36,8 +36,11 @@ import net.primal.shared.data.local.serialization.ListsTypeConverters
         LocalAppData::class,
         LocalAppSessionEventData::class,
     ],
-    version = 17,
+    version = 18,
     exportSchema = true,
+    autoMigrations = [
+        androidx.room.AutoMigration(from = 17, to = 18),
```
AleksandarIlic: Use import.

app/src/main/kotlin/net/primal/android/signer/provider/AppDisplayInfo.kt
```diff
@@ -11,9 +11,9 @@ data class AppDisplayInfo(
 )
 
 @Composable
-fun rememberAppDisplayInfo(packageName: String): AppDisplayInfo {
+fun rememberAppDisplayInfo(packageName: String, fallbackAppName: String? = null): AppDisplayInfo {
```
AleksandarIlic: We should provide fallback here. We can always use `appDisplayInfo.name ?: fallbackName`

data/account/local/src/commonMain/kotlin/net/primal/data/account/local/dao/apps/local/LocalAppData.kt
```diff
@@ -8,6 +8,7 @@ import net.primal.data.account.local.dao.apps.TrustLevel
 data class LocalAppData(
     @PrimaryKey val identifier: String,
     val packageName: String,
+    val name: String? = null,
```
AleksandarIlic: nitpick: Put fields with default values after mandatory fields. Not sure if we are following this with db tables entities.

PR #837, opened 2026-01-06
Prevent session event logging for unknown local apps

markocic:


AleksandarIlic (changes requested):
Let's check if app exists on entry levels.
- in content provider
- and in activity for intents which are not read_public_key

we can resolve app identifier from `LocalSignerMethod`

PR #843, opened 2026-01-13
Fix reply published as root note on low connectivity

mehmedalijaK:


app/src/main/kotlin/net/primal/android/editor/NoteEditorViewModel.kt
```diff
@@ -548,14 +550,31 @@ class NoteEditorViewModel @AssistedInject constructor(
...
+
+                    val rootNoteNevent = rootPostId?.let { id ->
+                        val localRoot = feedRepository.findPostsById(id)
+                        localRoot?.asFeedPostUi()?.asNevent()
+                            ?: Nevent(
+                                kind = NostrEventKind.ShortTextNote.value,
+                                userId = null,
+                                eventId = id,
+                            )
+                    }
```
markocic: We can resolve root nevent purely from reply to note. Root marker tags have `eventId` and `userId` so there is no need for the extra query of the database. If there is no root marker, `replyNevent` and `rootNevent` should be the same.

PR #853, opened 2026-01-20
Implement recovery phrase validation with debounce

mehmedalijaK:


app/src/main/kotlin/net/primal/android/wallet/di/WalletUtilitiesModule.kt
```diff
@@ -44,4 +45,8 @@ object WalletUtilitiesModule {
             primalWalletApiClient = primalApiClient,
             nostrEventSignatureHandler = nostrNotary,
         )
+
+    @Provides
+    @Singleton
+    fun providesRecoveryPhraseValidator(): RecoveryPhraseValidator = RecoveryPhraseValidator()
```
AleksandarIlic: No need to inject this simple class without any dependencies. Just instantiate it where you use it.

AleksandarIlic (approved):

PR #859, opened 2026-01-23
Implement NwcWalletClient for NIP-47

mehmedalijaK:


data/wallet/remote-nwc/src/commonMain/kotlin/net/primal/core/networking/nwc/wallet/NwcWalletClient.kt
```diff
@@ -0,0 +1,159 @@
...
+    private val _errors: Channel<WalletNwcRequestException> = Channel(Channel.BUFFERED)
+    val errors = _errors.receiveAsFlow()
+
+    suspend fun connect() =
+        runCatching {
+            nostrSocketClient.ensureSocketConnectionOrThrow()
+            startSubscriptionListener()
+        }
+
+    suspend fun updateConnections(connections: List<NwcConnection>) {
```
AleksandarIlic, line 77: Do you really need the entire `NwcConnection` or only service pubkey?

data/wallet/remote-nwc/src/commonMain/kotlin/net/primal/core/networking/nwc/wallet/NwcWalletClient.kt
```diff
@@ -0,0 +1,159 @@
...
+            connectionMap = connections.associateBy { it.serviceKeyPair.pubKey }
+            connectionMap.keys.toList()
+        }
+
+        if (servicePubKeys.isNotEmpty()) {
+            sendFilterRequest(servicePubKeys)
+        }
+    }
+
+    private suspend fun sendFilterRequest(secretPubKeys: List<String>) {
```
AleksandarIlic: You named the argument here `secretPubKeys`, but when you call this you pass `servicePubKeys`. Not sure if those are the same, but I think we should have name consistency.

data/wallet/remote-nwc/src/commonMain/kotlin/net/primal/core/networking/nwc/wallet/NwcWalletClient.kt
```diff
@@ -0,0 +1,159 @@
...
+    private val nostrSocketClient = NostrSocketClientFactory.create(
+        wssUrl = relayUrl,
+        onSocketConnectionOpened = { url ->
+            onSocketConnectionOpened?.invoke(url)
+            scope.launch {
+                val secretPubKeys = connectionMapMutex.withLock { connectionMap.keys.toList() }
+                if (secretPubKeys.isNotEmpty()) {
+                    sendFilterRequest(secretPubKeys)
+                }
+            }
```
AleksandarIlic, line 61: We should not start the connection automatically as soon as the object is created. There is an explicit `connect()` function which can be called to initiate the connection and start listening.

data/wallet/remote-nwc/src/commonMain/kotlin/net/primal/core/networking/nwc/wallet/NwcWalletClient.kt
```diff
@@ -0,0 +1,159 @@
...
+                    put(
+                        "#p",
+                        buildJsonArray {
+                            secretPubKeys.forEach { add(it) }
+                        },
+                    )
+                },
+            )
+        }.onFailure {
+            Napier.w(tag = "NwcWalletClient", throwable = it) { "Failed to update NWC subscription." }
```
AleksandarIlic: I think we should not silently fail here. In case `sendREQ` fails, the callers needs to know that. 

I assume `updateConnections` and `connect` are function which would throw this and the callers must handle this gracefully. So, better to remove `runCatching`.

data/wallet/remote-nwc/src/commonMain/kotlin/net/primal/core/networking/nwc/wallet/NwcWalletClient.kt
```diff
@@ -0,0 +1,159 @@
...
+
+        if (servicePubKeys.isNotEmpty()) {
+            sendFilterRequest(servicePubKeys)
+        }
+    }
+
+    private suspend fun sendFilterRequest(secretPubKeys: List<String>) {
+        runCatching {
+            nostrSocketClient.sendREQ(
+                subscriptionId = subscriptionId,
```
AleksandarIlic: If this function fails, are you sure we are good to use the same `subscriptionId` again? Maybe we should create new subscription id per call, just to be sure.

data/wallet/remote-nwc/src/commonMain/kotlin/net/primal/core/networking/nwc/wallet/NwcWalletClient.kt
```diff
@@ -0,0 +1,159 @@
...
+            )
+        }
+    }
+
+    suspend fun destroy() {
+        nostrSocketClient.close()
+        scope.cancel()
+    }
+
+    suspend fun publishEvent(event: NostrEvent): Result<Unit> =
```
AleksandarIlic: `connect` and `updateConnections` are without `Result`. We should be consistent. Since this is a networking class, maybe we should throw exceptions.

markocic (approved):

PR #870, opened 2026-02-02
Implement CSV export for wallet transactions

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/utils/CsvUtils.kt
```diff
@@ -0,0 +1,107 @@
...
+import java.io.BufferedWriter
+import java.io.OutputStreamWriter
+import java.io.Writer
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import net.primal.domain.nostr.toNostrString
+import net.primal.domain.profile.ProfileData
+import net.primal.domain.transactions.Transaction
+
+private val CSV_HEADERS = listOf(
```
AleksandarIlic: This file does not belong to `core.utils`, because it's not a general utility class. It's something for the wallet - you should move it to the `wallet` package and name it properly, e.g. `TransactionsCsvExporter` or `TransactionsCsvUtils`

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsContract.kt
```diff
@@ -16,11 +17,15 @@ interface WalletSettingsContract {
         val activeAccountLegendaryCustomization: LegendaryCustomization? = null,
         val activeAccountBlossoms: List<String> = emptyList(),
         val activeAccountDisplayName: String = "",
+        val isExportingTransactions: Boolean = false,
+        val transactions: List<Transaction> = emptyList(),
```
AleksandarIlic: Instead of `transactions` you have to use SideEffect. Side effects are launched whenever something is happening that affects the env. In this case, prepared transactions are affecting env since it must launch a share dialog. It's not internal state.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsScreen.kt
```diff
@@ -87,6 +95,25 @@ fun WalletSettingsScreen(
...
+            }
+        }
+        viewModel.setEvent(UiEvent.TransactionsExported)
+    }
+
+    LaunchedEffect(uiState.value.transactions) {
+        if (uiState.value.transactions.isNotEmpty()) {
+            saveFileLauncher.launch("PrimalTransactions.csv")
+        }
+    }
```
AleksandarIlic: Please handle this with side effect from view model.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsScreen.kt
```diff
@@ -87,6 +96,34 @@ fun WalletSettingsScreen(
         }
     }
 
+    val context = LocalContext.current
+    val scope = rememberCoroutineScope()
+    var pendingTransactions by remember { mutableStateOf<List<Transaction>>(emptyList()) }
```
markocic: Can we lose this state variable? It is used to transfer data between but we can do it directly as an argument.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsScreen.kt
```diff
@@ -87,6 +96,34 @@ fun WalletSettingsScreen(
...
+            pendingTransactions = emptyList()
+        }
+    }
+
+    LaunchedEffect(viewModel) {
+        viewModel.effects.collect { effect ->
+            when (effect) {
+                is WalletSettingsContract.SideEffect.ExportTransactions -> {
+                    pendingTransactions = effect.transactions
+                    saveFileLauncher.launch("PrimalTransactions.csv")
```
markocic: It might be better for name to not be hard-coded. User might want to export multiple wallets and this would be annoying. Name could be derived from wallet.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsScreen.kt
```diff
@@ -179,7 +216,19 @@ fun WalletSettingsScreen(
...
+                            CircularProgressIndicator(
+                                modifier = Modifier.size(24.dp),
+                                strokeWidth = 2.dp,
+                            )
+                        } else {
+                            Icon(imageVector = PrimalIcons.DownloadsFilled, contentDescription = null)
+                        }
+                    },
+                    onClick = {
+                        if (!state.isExportingTransactions) {
```
markocic: Could we implement `disabled` state for `SettingsItem` and just disable it while exporting? Export could take a while and having visual feedback would be nice. Disabled state could just be reducing opacity - VISUALLY. Don't actually change alpha value of whole item, pick slightly lighter/darker colors.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsContract.kt
```diff
@@ -16,17 +17,23 @@ interface WalletSettingsContract {
...
         data object RequestFetchWalletConnections : UiEvent()
+        data object ExportTransactions : UiEvent()
         data class RevokeConnection(val nwcPubkey: String) : UiEvent()
         data class UpdateUseExternalWallet(val value: Boolean) : UiEvent()
         data class UpdateMinTransactionAmount(val amountInSats: Long) : UiEvent()
         data class ConnectExternalWallet(val connectionLink: String) : UiEvent()
     }
 
+    sealed class SideEffect {
+        data class ExportTransactions(val transactions: List<Transaction>) : SideEffect()
```
markocic: I don't like the idea of having same names for `UiEvent` and `SideEffect`. I am confused as to who is actually exporting and what is going on. Let's go with better naming to reflect what is actually going on.

markocic (approved):

PR #872, opened 2026-02-02
Implement sending nwc info event on relay when connection is established

markocic:


data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/service/NwcServiceImpl.kt
```diff
@@ -104,13 +109,44 @@ class NwcServiceImpl internal constructor(
...
                         }
+                    publishInfoEventsIfNeeded(connections)
                 } else {
                     Napier.d(tag = TAG) { "No connections found, disconnecting from relay" }
                     disconnectFromRelay()
                 }
             }
         }
 
+    private suspend fun publishInfoEventsIfNeeded(connections: List<NwcConnection>) {
```
AleksandarIlic: `ensureInfoEventArePublished` might be a better name.

AleksandarIlic (approved):

PR #879, opened 2026-02-05
Implement `lookup_invoice` nwc method

markocic:


data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/handler/TransactionsHandler.kt
```diff
@@ -37,10 +42,36 @@ internal class TransactionsHandler(
                 walletDatabase.walletTransactions().upsertAll(
                     data = transactions.map { it.toWalletTransactionData() },
                 )
+
+                // Update NwcInvoice state for settled incoming transactions
+                val settledTransactions = transactions.filter { tx ->
+                    tx.state == TxState.SUCCEEDED &&
+                        tx.type == TxType.DEPOSIT &&
+                        tx.extractPaymentHash() != null
```
AleksandarIlic: `val paymentHash = tx.extractPaymentHash() ?: continue`

You already have this in the for loop. Do we really need to extract it twice? Maybe just skip it in the filter?

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/handler/TransactionsHandler.kt
```diff
@@ -37,10 +42,36 @@ internal class TransactionsHandler(
...
+
+                for (tx in settledTransactions) {
+                    val paymentHash = tx.extractPaymentHash() ?: continue
+                    val settledAt = tx.completedAt ?: kotlin.time.Clock.System.now().epochSeconds
+                    val preimage = tx.extractPreimage()
+
+                    Napier.d(tag = TAG) {
+                        "Marking NwcInvoice as settled: paymentHash=$paymentHash, settledAt=$settledAt"
+                    }
+                    walletDatabase.nwcInvoices().markSettledWithDetails(
```
AleksandarIlic: Maybe wrap the for loop in wallet transaction?

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/processor/NwcRequestProcessor.kt
```diff
@@ -251,14 +256,36 @@ class NwcRequestProcessor internal constructor(
...
+                        createdAt = nowSeconds,
+                        expiresAt = expiresAtSeconds ?: (nowSeconds + DEFAULT_INVOICE_EXPIRY_SECONDS),
+                        settledAt = null,
+                        preimage = null,
+                        state = NwcInvoiceState.PENDING,
+                    ),
+                )
+                Napier.d(tag = TAG) { "MakeInvoice: persisted NwcInvoice for lookup" }
+            }.onFailure { e ->
+                Napier.e(tag = TAG, throwable = e) { "MakeInvoice: failed to persist NwcInvoice" }
```
AleksandarIlic, line 290: Should we let REQ completes successfully if we can not save it in the db?

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/processor/NwcRequestProcessor.kt
```diff
@@ -307,17 +335,77 @@ class NwcRequestProcessor internal constructor(
...
+        if (nwcInvoice != null) {
+            Napier.d(tag = TAG) { "LookupInvoice: found NwcInvoice, state=${nwcInvoice.state}" }
+
+            val nowSeconds = Clock.System.now().epochSeconds
+            val effectiveState = when {
+                nwcInvoice.state == NwcInvoiceState.SETTLED -> "settled"
+                nwcInvoice.state == NwcInvoiceState.EXPIRED -> "expired"
+                nwcInvoice.expiresAt < nowSeconds -> {
+                    runCatching {
+                        walletRepository.markNwcInvoiceExpired(nwcInvoice.invoice)
```
AleksandarIlic: Are INVOICES in our DB stay in PENDING state until someone asks for them? This might be a misleading state. Someone might just query the db and think it's still PENDING.

Maybe the state should be dynamically calculated instead of persisted as a column?

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/processor/NwcRequestProcessor.kt
markocic: I think it should. The endpoint is for making invoice and we have made an invoice. Failure to persist to DB would only cause lookup_invoice to return not found until it is settled. As we have concluded that this local cache is a workaround I don't think we should fail the whole method.

AleksandarIlic: Sometimes workarounds stay forever. 😄

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/processor/NwcRequestProcessor.kt
```diff
@@ -251,14 +256,36 @@ class NwcRequestProcessor internal constructor(
                 lnbc = invoice,
             )
```
AleksandarIlic: You should handle this gracefully if this throws exception.

PR #882, opened 2026-02-06
Implement periodic cleanup for expired budget holds

markocic:


data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/service/NwcServiceImpl.kt
```diff
@@ -29,9 +31,11 @@ import net.primal.domain.connections.nostr.model.NwcConnection
 import net.primal.domain.nostr.cryptography.utils.unwrapOrThrow
 import net.primal.wallet.data.nwc.NwcCapabilities
 import net.primal.wallet.data.nwc.builder.NwcWalletResponseBuilder
+import net.primal.wallet.data.nwc.manager.NwcBudgetManager
 import net.primal.wallet.data.nwc.processor.NwcRequestProcessor
 
 private const val MAX_CACHE_SIZE = 20
+private val CLEANUP_INTERVAL = 2.minutes
```
AleksandarIlic: Why don't we use `companion object` for constants?

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/service/NwcServiceImpl.kt
```diff
@@ -54,10 +59,21 @@ class NwcServiceImpl internal constructor(
...
+        startPeriodicCleanup()
         observeConnections(userId)
         observeRetrySendResponseQueue()
     }
 
+    private fun startPeriodicCleanup() {
+        scope.launch {
+            while (true) {
+                Napier.d(tag = TAG) { "Running cleanupExpiredHolds" }
+                budgetManager.cleanupExpiredHolds()
```
AleksandarIlic: Maybe wrap this in `runCatching`? Just to be sure if something goes to be able to retry again in next round. I see that we are good with current code, but if code changes in the future, then we could see some issues.

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/service/NwcServiceImpl.kt
```diff
@@ -54,10 +59,21 @@ class NwcServiceImpl internal constructor(
...
     override fun initialize(userId: String) {
         Napier.d(tag = TAG) { "NwcService initializing for userId=$userId" }
+        startPeriodicCleanup()
         observeConnections(userId)
         observeRetrySendResponseQueue()
     }
 
+    private fun startPeriodicCleanup() {
+        scope.launch {
+            while (true) {
```
AleksandarIlic: Maybe we should check if `scope.isActive` on every loop?

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/processor/NwcRequestProcessor.kt
```diff
@@ -391,7 +391,7 @@ class NwcRequestProcessor internal constructor(
     }
 
     companion object {
-        private const val PAYMENT_HOLD_TIMEOUT_MS = 60_000L
+        private const val PAYMENT_HOLD_TIMEOUT_MS = 300_000L // 5 minutes
```
AleksandarIlic: Can we use `5.minutes.inWholeMilliseconds`? Here and on other places for durations. It's much easier to read for humans.

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/nwc/service/NwcServiceImpl.kt
AleksandarIlic: And 2 minutes is maybe to big number. We should run this more frequently. Maybe `30.seconds`. Why? Because the hold is 5.minutes and if it happens just after cleaning, it's going to take 7.minutes.

AleksandarIlic (changes requested):
Great work. Minor improvements in comments.

PR #883, opened 2026-02-06
Support NWC connections for both wallets

mehmedalijaK:


app/src/main/kotlin/net/primal/android/settings/wallet/nwc/primal/create/CreateNewWalletConnectionViewModel.kt
```diff
@@ -66,23 +71,35 @@ class CreateNewWalletConnectionViewModel @Inject constructor(
...
-                        .toPlainString()
+                val userId = activeAccountStore.activeUserId()
+                val nwcConnectionUri = when (val activeWallet = walletAccountRepository.getActiveWallet(userId)) {
+                    is Wallet.Spark -> {
+                        nwcRepository.createNewWalletConnection(
+                            userId = userId,
+                            walletId = activeWallet.walletId,
+                            appName = appName,
+                            dailyBudget = dailyBudget,
+                        ).getOrNull()
```
markocic: We are not handling errors from this function. We are effectively ignoring them.

app/src/main/kotlin/net/primal/android/settings/wallet/nwc/primal/create/CreateNewWalletConnectionViewModel.kt
```diff
@@ -66,23 +71,35 @@ class CreateNewWalletConnectionViewModel @Inject constructor(
...
-                    dailyBudget = formattedDailyBudgetBtc,
-                )
-
                 setState {
                     copy(
-                        nwcConnectionUri = response.nwcConnectionUri,
+                        nwcConnectionUri = nwcConnectionUri,
                         creatingSecret = false,
                     )
                 }
```
markocic, line 102: Switch to `runCatching` as `catch` block is the same for both cases with a minor message change that could be seen from `error` variable.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsScreen.kt
```diff
@@ -264,23 +261,11 @@ fun WalletSettingsScreen(
                     onCreateNewWalletConnection = onCreateNewWalletConnection,
                     connectionsState = state.connectionsState,
                     onRetryFetchingConnections = { eventPublisher(UiEvent.RequestFetchWalletConnections) },
-                    isPrimalWalletActivated = state.activeWallet?.isPrimalWalletAndActivated() == true,
+                    walletSupportsNwcConnections = state.activeWallet.supportsNwcConnections(),
                 )
 
                 Spacer(modifier = Modifier.height(16.dp))
 
-                SettingsItem(
```
markocic, line 287: Can you also delete the screen and the navigate from developer tools?

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -132,11 +153,20 @@ class WalletSettingsViewModel @AssistedInject constructor(
...
+                when (activeWallet) {
+                    is Wallet.Spark -> {
+                        nwcRepository.revokeConnection(activeAccountStore.activeUserId(), nwcPubkey)
+                    }
+                    is Wallet.Primal -> {
+                        primalWalletNwcRepository.revokeConnection(activeAccountStore.activeUserId(), nwcPubkey)
+                    }
+                    else -> Unit
+                }
             } catch (error: SignatureException) {
```
markocic: Go with `runCatching` and revert connections on failure and log the error.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -132,11 +153,20 @@ class WalletSettingsViewModel @AssistedInject constructor(
 
     private fun revokeNwcConnection(nwcPubkey: String) =
         viewModelScope.launch {
+            val activeWallet = state.value.activeWallet
             val nwcConnections = state.value.nwcConnectionsInfo
             try {
                 val updatedConnections = nwcConnections.filterNot { it.nwcPubkey == nwcPubkey }
```
markocic: We are working with `PrimalNwcConnectionInfo` domain model but these are both primal and native wallet connections. Let's make a UiModel so we are not dependant on the domain model that might be deleted tomorrow as we are in the process of deprecation.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -83,9 +89,24 @@ class WalletSettingsViewModel @AssistedInject constructor(
 
     private fun fetchWalletConnections() =
         viewModelScope.launch {
+            val activeWallet = state.value.activeWallet
+            if (!activeWallet.supportsNwcConnections()) {
```
markocic: During initialization `activeWallet` will be `null` as nothing will guarantee us that active wallet observers will fill state in time. This will likely cause a premature Error state in some cases. When I go into `WalletSettingsScreen` I can see the error message first and then on retry it says that I don't have any connections. Please resolve this issue.

domain/wallet/src/commonMain/kotlin/net/primal/domain/utils/WalletUtils.kt
```diff
@@ -6,3 +6,10 @@ import net.primal.domain.wallet.WalletKycLevel
 fun Wallet.isPrimalWalletAndActivated() = this is Wallet.Primal && this.kycLevel != WalletKycLevel.None
 
 fun Wallet?.isConfigured(): Boolean = this != null && (this !is Wallet.Primal || this.kycLevel != WalletKycLevel.None)
+
+fun Wallet?.supportsNwcConnections(): Boolean =
```
markocic: Should this go to `WalletCapabilities`? Wallet.NWC doesn't support but other ones do. Primal is just dependant on other state but generally it does support nwc. I think this should be changed.

app/detekt-baseline.xml
```diff
@@ -99,6 +99,7 @@
...
     <ID>LongParameterList:WalletDashboardViewModel.kt$WalletDashboardViewModel$( userRepository: UserRepository, private val activeAccountStore: ActiveAccountStore, private val walletAccountRepository: WalletAccountRepository, private val walletRepository: WalletRepository, private val primalBillingClient: PrimalBillingClient, private val billingRepository: BillingRepository, private val subscriptionsManager: SubscriptionsManager, private val exchangeRateHandler: ExchangeRateHandler, private val ensureSparkWalletExistsUseCase: EnsureSparkWalletExistsUseCase, private val sparkWalletAccountRepository: SparkWalletAccountRepository, )</ID>
+    <ID>LongParameterList:WalletSettingsViewModel.kt$WalletSettingsViewModel$( @Assisted private val nwcConnectionUrl: String?, private val activeAccountStore: ActiveAccountStore, private val walletRepository: WalletRepository, private val walletAccountRepository: WalletAccountRepository, private val primalWalletNwcRepository: PrimalWalletNwcRepository, private val nwcRepository: NwcRepository, private val connectNwcUseCase: ConnectNwcUseCase, )</ID>
```
markocic: How do we have this entry if there is `@Suppress` on `WalletSettingsViewModel` for this exact issue?

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -316,3 +346,10 @@ class WalletSettingsViewModel @AssistedInject constructor(
             setState { copy(isRevertingToPrimalWallet = false, showRevertToPrimalWallet = false) }
         }
 }
+
+private fun NwcConnection.asPrimalNwcConnectionInfo() =
```
markocic: Let's not map one domain object into another.

app/src/main/kotlin/net/primal/android/settings/wallet/nwc/primal/create/CreateNewWalletConnectionViewModel.kt
```diff
@@ -63,35 +66,43 @@ class CreateNewWalletConnectionViewModel @Inject constructor(
...
+                                .stripTrailingZeros()
+                                .toPlainString()
+                        }
+                        primalWalletNwcRepository.createNewWalletConnection(
+                            userId = userId,
+                            appName = appName,
+                            dailyBudget = formattedDailyBudgetBtc,
+                        ).nwcConnectionUri
+                    }
+                    else -> null
```
markocic: We should throw here as other wallets don't support nwc so we go into `onFailure`. Right now we can't get to this screen with other wallets but that might change in the future and we then we have to remember this case or risk introducing a bug.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/ui/model/WalletNwcConnectionUi.kt
```diff
@@ -0,0 +1,25 @@
+package net.primal.android.settings.wallet.settings.ui.model
+
+import net.primal.core.utils.CurrencyConversionUtils.toBtc
+import net.primal.domain.connections.nostr.model.NwcConnection
+import net.primal.domain.connections.primal.model.PrimalNwcConnectionInfo
+
+data class WalletNwcConnectionUi(
+    val appName: String,
+    val dailyBudget: String?,
```
markocic: change to `dailyBudgetInBtc` so it's clear.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -81,21 +82,36 @@ class WalletSettingsViewModel @AssistedInject constructor(
...
 
-    private fun fetchWalletConnections() =
+    private fun fetchWalletConnections(wallet: Wallet? = null) =
         viewModelScope.launch {
-            try {
-                setState { copy(connectionsState = WalletSettingsContract.ConnectionsState.Loading) }
-                val connections = primalWalletNwcRepository.getConnections(userId = activeAccountStore.activeUserId())
+            val activeWallet = wallet ?: state.value.activeWallet ?: return@launch
+            if (!activeWallet.capabilities.supportsNwcConnections) {
+                setState { copy(connectionsState = WalletSettingsContract.ConnectionsState.Error) }
```
markocic, line 109: Let's see if we are going to hide the section. Right now this won't be good as we are showing that we can't fetch connections even tho the active wallet doesn't even support nwc. It's misleading.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -81,21 +82,36 @@ class WalletSettingsViewModel @AssistedInject constructor(
         observeEvents()
     }
 
-    private fun fetchWalletConnections() =
+    private fun fetchWalletConnections(wallet: Wallet? = null) =
```
markocic: I just realized that we are not actually fetching connections for internal wallet as they are in our local db. Let's split the logic.

For internal wallet: we should have observe on connections, no optimistic updates
For primal wallet: we should have this fetch connections with optimistic updates and UiEvent to request updates

Right now this looks like it's making internal wallet logic fit primal wallets logic but it's a whole different case. I totally missed this the first time I was reviewing this.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -81,21 +82,36 @@ class WalletSettingsViewModel @AssistedInject constructor(
...
         viewModelScope.launch {
-            try {
-                setState { copy(connectionsState = WalletSettingsContract.ConnectionsState.Loading) }
-                val connections = primalWalletNwcRepository.getConnections(userId = activeAccountStore.activeUserId())
+            val activeWallet = wallet ?: state.value.activeWallet ?: return@launch
+            if (!activeWallet.capabilities.supportsNwcConnections) {
```
markocic: Should we check for kyc for primal wallet like you did in previous iteration? I see that we are losing this check now so wondering why?

data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/repository/NwcRepositoryImpl.kt
```diff
@@ -63,6 +63,11 @@ internal class NwcRepositoryImpl(
             ).asSuccess()
         }
 
+    override suspend fun revokeConnection(userId: String, secretPubKey: String) =
+        with(dispatcherProvider.io()) {
```
AleksandarIlic: `withContext` must be used. Please fix all cases in this class and check if we are not making any everywhere else this mistakes.

AleksandarIlic (changes requested):
Partial review. I have yet to look `WalletSettingsViewModel` tomorrow.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -81,21 +86,40 @@ class WalletSettingsViewModel @AssistedInject constructor(
...
+                    setState {
+                        copy(
+                            nwcConnectionsInfo = connections.map { it.asWalletNwcConnectionUi() },
+                            connectionsState = WalletSettingsContract.ConnectionsState.Loaded,
+                        )
+                    }
+                }
+        }
+
+    private fun fetchPrimalWalletConnections(wallet: Wallet.Primal? = null): Job =
```
AleksandarIlic: If you are accepting wallet from arguments, you should then never fallback to the state. Make this mandatory argument and the callers should ensure that wallet is available before calling this.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsScreen.kt
```diff
@@ -258,29 +256,20 @@ fun WalletSettingsScreen(
...
-                    isPrimalWalletActivated = state.activeWallet?.isPrimalWalletAndActivated() == true,
-                )
+                if (state.activeWallet?.capabilities?.supportsNwcConnections == true) {
+                    ConnectedAppsSettings(
+                        primalNwcConnectionInfos = state.nwcConnectionsInfo,
+                        onRevokeConnectedApp = { eventPublisher(UiEvent.RevokeConnection(it)) },
+                        onCreateNewWalletConnection = onCreateNewWalletConnection,
+                        connectionsState = state.connectionsState,
+                        onRetryFetchingConnections = { eventPublisher(UiEvent.RequestFetchWalletConnections) },
+                        isWalletActivated = state.activeWallet is Wallet.Spark ||
```
AleksandarIlic: Remove `isWalletActivated`. We don't want to show any hints of activating Primal custodial wallet because we have removed activation flow. For self-custodial wallets this no longer applies.

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -170,6 +210,13 @@ class WalletSettingsViewModel @AssistedInject constructor(
                         )
                     }
 
+                    connectionsJob?.cancel()
+                    connectionsJob = when (wallet) {
```
AleksandarIlic: FYI: We are mixing here observing and fetching. When dealing with different implementations, we should always try to have the same interface on top.

Since this just a temporary solution, it can stay like this. We do not want to invest time into something that we know it's going to change in the future.

PR #884, opened 2026-02-06
Update wallet upgrade design and implement notice debounce

markocic:


app/src/main/kotlin/net/primal/android/user/domain/UserAccount.kt
```diff
@@ -51,22 +51,20 @@ data class UserAccount(
     val blossomServers: List<String> = emptyList(),
     val pushNotificationsEnabled: Boolean = false,
     val shouldShowStreamControlPopup: Boolean = true,
-    val shouldShowUpgradeWalletSheet: Boolean = true,
+    val lastUpgradeNoticeShownAtMillis: Long? = null,
```
AleksandarIlic: Can we avoid persisting anything related to this notice? If the app is restarting again we should show the notice if user has not yet upgraded. Between two bottomSheet notices we can keep the timestamps in memory only somewhere. Isn't the `UpgradeWalletSheetViewModel` lives for the entire time of the app?

app/src/main/kotlin/net/primal/android/wallet/upgrade/sheet/UpgradeWalletSheetContract.kt
```diff
@@ -5,7 +5,8 @@ import net.primal.domain.links.CdnImage
 
 interface UpgradeWalletSheetContract {
     data class UiState(
-        val shouldShowUpgradeNotice: Boolean = false,
+        val shouldUserUpgrade: Boolean = false,
+        val noticeDebouncePassed: Boolean = false,
```
AleksandarIlic: `shouldShowNotice` or `shouldWarnUser` is better name. debounce passed is impl. detail.

AleksandarIlic (changes requested):

PR #894, opened 2026-02-11
Route NWC connection by active wallet type and unify notification channel

mehmedalijaK:


app/src/main/kotlin/net/primal/android/nostrconnect/connect/NostrConnectContract.kt
```diff
@@ -30,6 +30,10 @@ interface NostrConnectContract {
     }
 
     sealed class SideEffect {
-        data class ConnectionSuccess(val callbackUri: String?) : SideEffect()
+        data class ConnectionSuccess(
+            val callbackUri: String?,
+            val userId: String,
+            val shouldStartNwcService: Boolean = false,
```
AleksandarIlic: Maybe `requiresNwcService` instead of `shouldStart`. Tells what it's needed, not what has to be done. The other side is going to decide what to do.

AleksandarIlic (changes requested):
LGTM. Just minor nitpick.

app/src/main/kotlin/net/primal/android/core/service/PrimalNwcService.kt
```diff
@@ -50,7 +50,7 @@ class PrimalNwcService : Service() {
     lateinit var accountsStore: UserAccountsStore
 
     companion object {
-        const val CHANNEL_ID = "nwc_wallet_service"
+        const val CHANNEL_ID = "remote_signer"
```
markocic: nitpick: We should extract this into a const that will be used in `PrimalRemoteSignerService` and here. Avoids possible typos, repetitions or accidental bugs due to different channel ids.

markocic (changes requested):

app/src/main/kotlin/net/primal/android/core/service/PrimalNwcService.kt
```diff
@@ -183,7 +183,7 @@ class PrimalNwcService : Service() {
         notificationManager.createNotificationChannel(
             NotificationChannel(
                 CHANNEL_ID,
-                getString(R.string.nwc_service_notification_channel_name),
+                getString(R.string.signer_notification_name),
```
markocic, line 185: Should we change channel name to be something like "Primal Service" or something like that? User will see "Remote Signer" as channel for both Remote Signer Service and Nwc Service.

markocic (approved):
LGTM

AleksandarIlic (approved):

PR #897, opened 2026-02-13
Implement NWC service auto-start feature using backend notifications

markocic:


app/src/google/kotlin/net/primal/android/core/push/PrimalFirebaseMessagingService.kt
```diff
@@ -20,11 +21,16 @@ class PrimalFirebaseMessagingService : FirebaseMessagingService() {
...
+    @Inject
+    lateinit var nwcRemoteMessageHandler: NwcRemoteMessageHandler
+
     val scope = CoroutineScope(SupervisorJob())
 
     override fun onMessageReceived(message: RemoteMessage) {
         if (signerRemoteMessageHandler.isRemoteSignerMessage(message = message)) {
             scope.launch { signerRemoteMessageHandler.process(message = message) }
+        } else if (nwcRemoteMessageHandler.isNwcMessage(message = message)) {
+            scope.launch { nwcRemoteMessageHandler.process(message = message) }
```
AleksandarIlic: Shouldn't we run this coroutines on IO? It seems that we are launching them on main threads...  If yes, we can create `scope` as `CoroutineScope(dispatcherProvider.io() + SupervisorJob())`

Also, both GlobalScope below and this `scope` are never canceled. Maybe we should add `withTimeout` to every call so it gets canceled with 15.seconds or so if something goes wrong here.

app/src/main/kotlin/net/primal/android/core/service/PrimalNwcService.kt
```diff
@@ -49,6 +52,9 @@ class PrimalNwcService : Service() {
     @Inject
     lateinit var accountsStore: UserAccountsStore
 
+    private val serviceScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
AleksandarIlic: Why `.Default`?  Service runs on main thread by default. Not sure if you want it to be on main or not.

app/src/main/kotlin/net/primal/android/settings/connected/details/remote/RemoteAppDetailsViewModel.kt
```diff
@@ -158,7 +158,9 @@ class RemoteAppDetailsViewModel @Inject constructor(
             runCatching {
                 connectionRepository.updateConnectionAutoStart(clientPubKey, enabled)
             }.onSuccess {
-                runCatching { tokenUpdater.updateTokenForRemoteSigner() }
+                CoroutineScope(dispatcherProvider.io()).launch {
```
AleksandarIlic: This is the practice we used for remote-signer as well? We used new CoroutineScope without withTimeout there?

app/src/main/kotlin/net/primal/android/settings/wallet/settings/WalletSettingsViewModel.kt
```diff
@@ -263,6 +277,30 @@ class WalletSettingsViewModel @AssistedInject constructor(
...
+                    setState { copy(isAutoStartEnabled = isEnabled) }
+                }
+        }
+
+    private fun updateAutoStart(enabled: Boolean) =
+        viewModelScope.launch {
+            runCatching {
+                nwcRepository.updateAutoStartForUser(userId = activeAccountStore.activeUserId(), autoStart = enabled)
+            }.onSuccess {
+                CoroutineScope(dispatchers.io()).launch {
```
AleksandarIlic: Do we have to use new CoroutineScope for every token updater call? I think we used that only for calls happening on the end of screen lifecycle, so it does not gets canceled. But, this is during the regular screen use, right?

data/account/repository/src/commonMain/kotlin/net/primal/data/account/repository/repository/PushNotificationRepositoryImpl.kt
```diff
@@ -19,7 +20,17 @@ class PushNotificationRepositoryImpl(
 
     override suspend fun updateNotificationTokenForNip46(authorizationEvent: NostrEvent, token: String) {
         withContext(dispatchers.io()) {
-            pushNotificationApi.updateNotificationTokenForNip46(authorizationEvent, token)
+            pushNotificationApi.updateNotificationTokenForNip46OrNip47(listOf(authorizationEvent), token)
```
AleksandarIlic: Nip460? 

Just kidding... 😄 

Maybe `updateNotificationTokenForRemoteSigners` or something like that...

app/src/main/kotlin/net/primal/android/core/service/PrimalNwcService.kt
```diff
@@ -52,7 +52,7 @@ class PrimalNwcService : Service() {
     @Inject
     lateinit var accountsStore: UserAccountsStore
 
-    private val serviceScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+    private val serviceScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```
AleksandarIlic, line 55: Why not `dispatcherProvider.io()`?

PR #899, opened 2026-02-13
Cleanup user data on logout for wallet tables

mehmedalijaK:


data/wallet/local/src/commonMain/kotlin/net/primal/wallet/data/local/dao/nwc/NwcWalletRequestLog.kt
```diff
@@ -10,7 +10,7 @@ data class NwcWalletRequestLog(
     val eventId: String,
     val connectionId: Encryptable<String>,
     val walletId: Encryptable<String>,
```
markocic, line 12: I just noticed that this was a mistake. We won't be able to match connections or wallets with these records. This is unrelated with this PR but should be addressed before this hits production. Will be a headache to fix after.

PR #907, opened 2026-02-19
Call wallet repository latest transactions sync on pull to refresh in dashboard

markocic:


app/src/main/kotlin/net/primal/android/wallet/dashboard/WalletDashboardScreen.kt
```diff
@@ -338,8 +349,8 @@ fun WalletDashboardScreen(
             PrimalPullToRefreshBox(
                 isRefreshing = pullToRefreshing,
                 onRefresh = {
-                    pullToRefreshing = true
-                    pagingItems.refresh()
+                    updatePullToRefreshing(true)
```
AleksandarIlic: This is strange name. Child function can not tell what parent function should do, only what happened, and parent function decides what to do.

app/src/main/kotlin/net/primal/android/wallet/dashboard/WalletDashboardViewModel.kt
```diff
@@ -154,6 +164,7 @@ class WalletDashboardViewModel @Inject constructor(
                 .filterNotNull()
                 .collect { wallet ->
                     fetchWalletBalance(walletId = wallet.walletId)
+                    syncLatestTransactions(walletId = wallet.walletId)
```
AleksandarIlic: Do we really need to call this here?

PR #911, opened 2026-02-23
Bugfix/wallet zaps rendering

markocic:


data/wallet/repository/src/commonMain/kotlin/net/primal/wallet/data/zaps/DefaultWalletNostrZapper.kt
```diff
@@ -14,6 +15,7 @@ import net.primal.domain.wallet.WalletRepository
 class DefaultWalletNostrZapper(
     private val lightningPayHelper: LightningPayHelper,
     private val walletRepository: WalletRepository,
+    private val eventRepository: EventRepository,
```
AleksandarIlic: I think this class could be used on iOS and we don't have `EventRepository` there. It should be optional.

PR #914, opened 2026-02-24
Implement await payment on receive payment screen

markocic:


app/src/main/kotlin/net/primal/android/wallet/transactions/receive/ReceivePaymentViewModel.kt
```diff
@@ -166,6 +171,11 @@ class ReceivePaymentViewModel @Inject constructor(
                     ),
                 )
             }
+
+            val wallet = _state.value.activeWallet
+            if (wallet?.type == WalletType.SPARK) {
```
AleksandarIlic: This should be done through wallet capabilities. Something like `canAwaitLnInvoicePayment`

PR #922, opened 2026-02-27
Remove Sunset and Sunrise Primal Themes

markocic:


app/src/main/kotlin/net/primal/android/core/activity/PrimalActivity.kt
```diff
@@ -105,10 +105,10 @@ abstract class PrimalActivity : FragmentActivity() {
 
     private fun Bundle?.restoreOrDefaultPrimalTheme(): PrimalTheme {
         return when (this) {
-            null -> PrimalTheme.Sunset
+            null -> PrimalTheme.Midnight
             else -> {
-                val themeName = this.getString("primalTheme", PrimalTheme.Sunset.themeName)
-                PrimalTheme.valueOf(themeName = themeName) ?: PrimalTheme.Sunset
+                val themeName = this.getString("primalTheme", PrimalTheme.Midnight.themeName)
+                PrimalTheme.valueOf(themeName = themeName) ?: PrimalTheme.Midnight
```
AleksandarIlic, line 111: If previously the theme was Sunset then the fallback should be dark blue. If it was Sunrise, then it should be light blue. Is this how are we handling compatibility?

app/src/main/kotlin/net/primal/android/core/compose/PrimalLoadingSpinner.kt
```diff
@@ -17,20 +16,11 @@ import com.airbnb.lottie.compose.LottieCompositionSpec
...
 import net.primal.android.R
-import net.primal.android.core.activity.LocalPrimalTheme
-import net.primal.android.theme.domain.PrimalAccent
 
 @Composable
 fun PrimalLoadingSpinner(size: Dp = 96.dp, paddingValues: PaddingValues = PaddingValues(all = 0.dp)) {
-    val primalTheme = LocalPrimalTheme.current
-    val animationRawResId = remember(primalTheme) {
-        when (primalTheme.accent) {
-            PrimalAccent.Summer -> R.raw.primal_loading_spinner_summer
```
AleksandarIlic, line 28: We should remove this resources if this is not done.

app/src/main/kotlin/net/primal/android/core/activity/PrimalActivity.kt
markocic: Yes. That's handled in `PrimalTheme::valueOf`.

PR #923, opened 2026-03-02
Update wallet screen title and account creation button

mehmedalijaK:


app/src/main/kotlin/net/primal/android/wallet/dashboard/WalletDashboardScreen.kt
```diff
@@ -163,9 +163,9 @@ fun WalletDashboardScreen(
     val pagingItems = state.transactions.collectAsLazyPagingItems()
     val listState = pagingItems.rememberLazyListStatePagingWorkaround()
 
-    val canShowWalletPicker = state.userWallets.isNotEmpty() && state.userWallets.size > 1 && state.wallet != null
+    val canShowWalletPicker = false
```
markocic: Nitpick: could we add const that will let us enable/disable wallet picker just by flipping it? Rest of the checks should be preserved so we don't have to figure it out later or look through git history for it.

AleksandarIlic: This is actually a minor effort, but can be helpful in the future. Let's do this. It should be internal to this file.

PR #931, opened 2026-03-05
Handle poll publishing

markocic:


app/src/main/kotlin/net/primal/android/editor/domain/PollPublishRequest.kt
```diff
@@ -0,0 +1,9 @@
+package net.primal.android.editor.domain
+
+data class PollPublishRequest(
+    val isZapPoll: Boolean,
+    val choices: List<Pair<String, String>>,
```
AleksandarIlic: It's unclear what Pair represents. Could we have a small data class with two strings?

markocic: Fixed. Replaced `Pair<String, String>` with a `PollOption(id, label)` data class.

PR #936, opened 2026-03-09
Update feed specs to multi-kind format with in-place spec transformation

mehmedalijaK:


domain/primal/src/commonMain/kotlin/net/primal/domain/feeds/FeedExtensions.kt
```diff
@@ -2,12 +2,18 @@ package net.primal.domain.feeds
 
 import net.primal.domain.nostr.cryptography.utils.hexToNpubHrp
 
+fun String.migrateFeedSpec(): String {
+    return this
+        .replace("\"kind\":\"notes\"", "\"kinds\":[1]")
+        .replace("\"kind\":\"reads\"", "\"kinds\":[20,30023]")
```
AleksandarIlic, line 8: Remove `20`. It's not related with articles. It's picture note and it goes within Notes feeds.

domain/primal/src/commonMain/kotlin/net/primal/domain/feeds/FeedExtensions.kt
```diff
@@ -99,15 +105,15 @@ fun String.resolveFeedSpecKind(): FeedSpecKind? {
     }
 }
 
-fun String.isNotesFeedSpec() = this.contains("\"kind\":\"notes\"") || this.contains("kind:1")
+fun String.isNotesFeedSpec() = this.contains("\"kinds\":[1]") || this.contains("kind:1")
```
AleksandarIlic, line 108: We should be careful here - this logic actually checks if the feed spec is rendering feed posts (which can be notes, pictures, polls, etc).

domain/primal/src/commonMain/kotlin/net/primal/domain/feeds/FeedExtensions.kt
```diff
@@ -99,15 +105,15 @@ fun String.resolveFeedSpecKind(): FeedSpecKind? {
...
 
 fun String.isImageSpec() = this.contains("\"query\":\"filter:image")
 
 fun String.isVideoSpec() = this.contains("\"query\":\"filter:video")
 
 fun String.isAudioSpec() = this.contains("\"query\":\"filter:audio")
 
-fun String.isReadsFeedSpec() = this.contains("\"kind\":\"reads\"") || this.contains("kind:30023")
+fun String.isReadsFeedSpec() = this.contains("\"kinds\":[20,30023]") || this.contains("kind:30023")
```
AleksandarIlic, line 116: The same as above logic. Check if in `kinds` there is 30023.

PR #960, opened 2026-03-13
Update signer prompt label from 'handle' to 'allow'

mehmedalijaK:


AleksandarIlic:
We can't actually doing like this. It's more complicated.

PR #962, opened 2026-03-16
Fix note editor reading clipboard on every keystroke

mehmedalijaK:


app/src/main/kotlin/net/primal/android/editor/NoteEditorScreen.kt
```diff
@@ -656,8 +653,8 @@ private fun NoteEditorInputArea(
                 .onSizeChanged { onReplyNoteHeightChanged(it.height) },
             value = state.content,
             onValueChange = {
-                val clipboardText = clipboardManager.getText()
-                if (clipboardText != null && it.text.contains(clipboardText.text)) {
+                val charsDelta = it.text.length - state.content.text.length
+                if (charsDelta > 1) {
```
markocic, line 657: It's worth noting that `charsDelta` can be higher when using swipe to type, voice text or emojis. Considering that `PasteContent` handler will try to parse nostr uris out of it, I don't think it's much a concern. False positives will happen. Might be good to increase this to 20 characters to eliminate extra string parsing.

PR #968, opened 2026-03-17
Add NIP-89 client tag to specific event kinds

mehmedalijaK:


app/src/test/kotlin/net/primal/android/editor/NotePublishHandlerTest.kt
```diff
@@ -608,7 +610,7 @@ class NotePublishHandlerTest {
                     withArg {
                         it.pubKey shouldBe expectedUserId
                         it.content shouldBe expectedContent
-                        it.tags shouldBe expectedTags
+                        it.tags.filterNot { tag -> tag[0].jsonPrimitive.content == "client" } shouldBe expectedTags
```
markocic: Shouldn't the tests be updated to now require client tag instead of filtering it out? Goes for all instances.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/events/EventInteractionRepositoryImpl.kt
```diff
@@ -48,7 +52,8 @@ class EventInteractionRepositoryImpl(
                 unsignedNostrEvent = NostrUnsignedEvent(
                     pubKey = userId,
                     kind = NostrEventKind.Reaction.value,
-                    tags = listOf(eventId.asEventIdTag(), eventAuthorId.asPubkeyTag()) + optionalTags,
+                    tags = listOf(eventId.asEventIdTag(), eventAuthorId.asPubkeyTag()) + optionalTags +
```
markocic: nitpick: we can just add in original `listOf` instead of performing addition on lists. Goes for other places as well.

markocic (changes requested):

PR #972, opened 2026-03-18
Fix wallet upgrade notice shown after returning from background

mehmedalijaK:


app/src/main/kotlin/net/primal/android/wallet/notice/sheet/WalletNoticeSheetViewModel.kt
```diff
@@ -104,6 +104,27 @@ class WalletNoticeSheetViewModel @Inject constructor(
...
+                        userId = userId,
+                        status = status,
+                        userAccount = userAccount,
+                    )
+                    setState { copy(noticeType = noticeType, shouldShowNotice = false) }
+                    if (noticeType != null) {
+                        scheduleNoticeIfEligible()
+                    }
+                }
+        }
```
markocic: This looks like the same code in `observeActiveAccountId()`. Let's make that a function instead to avoid code duplication.

PR #994, opened 2026-03-25
Feature/nip 05 verification

markocic:


data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/nip05/Nip05VerificationServiceImpl.kt
```diff
@@ -0,0 +1,205 @@
...
+
+    private suspend fun performVerification(pubkey: String, internetIdentifier: String): VerificationResult {
+        val parts = internetIdentifier.split("@")
+        if (parts.size != 2) {
+            Napier.d(tag = TAG) { "Invalid identifier format: $internetIdentifier" }
+            return VerificationResult(status = Nip05VerificationStatus.FAILED)
+        }
+
+        val localPart = parts[0]
+        val domain = parts[1]
```
AleksandarIlic, line 227: We should do some domain regex validation to avoid potential injections. Something like:

```kotlin
private val VALID_DOMAIN = Regex("^[a-zA-Z0-9]([a-zA-Z0-9-]*[a-zA-Z0-9])?(\\.[a-zA-Z0-9]([a-zA-Z0-9-]*[a-zA-Z0-9])?)+$")
if (!VALID_DOMAIN.matches(domain)) {
  return VerificationResult(status = Nip05VerificationStatus.FAILED)
}
  ```

File: data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/factory/CommonRepositoryFactory.kt [hunk 2]
```diff
@@ -224,10 +231,33 @@ abstract class CommonRepositoryFactory {
         )
     }
 
+    fun createNip05VerificationService(): Nip05VerificationService {
+        val nip05HttpClient = HttpClientFactory.createHttpClient {
```
AleksandarIlic: Is there a better place to keep this? I think we already have some class for http clients somewhere which we could call from this place and hide the impl. details of the nip05 http client.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/factory/CommonRepositoryFactory.kt
```diff
@@ -224,10 +231,33 @@ abstract class CommonRepositoryFactory {
...
+                    json = Json {
+                        ignoreUnknownKeys = true
+                        isLenient = true
+                    },
+                )
+            }
+            install(HttpTimeout) {
+                requestTimeoutMillis = 10_000
+                connectTimeoutMillis = 10_000
+            }
```
AleksandarIlic: We are dealing here with the user-generated links and unverified servers. We should be more careful. Maybe we should put some response validator to validate against content length. Not sure what the default content length for this kind of response is, but we could limit this to 1-2MB.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/ReferencedNoteCard.kt
```diff
@@ -69,6 +69,7 @@ fun ReferencedNoteCard(
             authorLegendaryCustomization = data.authorLegendaryCustomization,
             authorInternetIdentifier = data.authorInternetIdentifier,
             authorBlossoms = data.authorBlossoms,
+            authorId = data.authorId,
```
AleksandarIlic, line 72: is `authorNip05Status` omitted on purpose here? There a couple of places where we do not propagate this info down the stream:
- ReferencedNoteCard.kt
- NoteEditorScreen.kt (ReplyToNote)
- TransactionDetailsScreen.kt

Also, Unmapped nip05 fields in mappers. The field exists on a target type but never set:
- EventZap.asEventZapUiModel(),
- ReferencedNote.asFeedPostUi(),
- Stream.asReferencedStream()

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/factory/CommonRepositoryFactory.kt
```diff
@@ -224,10 +231,33 @@ abstract class CommonRepositoryFactory {
         )
     }
 
+    fun createNip05VerificationService(): Nip05VerificationService {
+        val nip05HttpClient = HttpClientFactory.createHttpClient {
+            followRedirects = false
```
AleksandarIlic: Why is this disabled? If a server is behind CDN or something like that, this might not work.

app/src/main/kotlin/net/primal/android/core/compose/NostrUserText.kt
```diff
@@ -71,10 +74,19 @@ fun NostrUserText(
...
     internetIdentifierBadgeAlign: PlaceholderVerticalAlign = PlaceholderVerticalAlign.Center,
     legendaryCustomization: LegendaryCustomization? = null,
+    nip05Status: Nip05VerificationStatus? = null,
+    profileId: String? = null,
     annotatedStringPrefixBuilder: (AnnotatedString.Builder.() -> Unit)? = null,
     annotatedStringSuffixBuilder: (AnnotatedString.Builder.() -> Unit)? = null,
 ) {
-    val verifiedBadge = !internetIdentifier.isNullOrEmpty()
+    val nip05Service = LocalNip05VerificationService.current
+    LaunchedEffect(profileId, internetIdentifier) {
```
AleksandarIlic, line 84: If `nip05Status != null` do we still have to call this? This will be fired on every post in the feed.

app/src/main/kotlin/net/primal/android/core/compose/profile/model/UserProfileItemUi.kt
```diff
@@ -19,6 +20,7 @@ data class UserProfileItemUi(
     val isFollowed: Boolean? = null,
     val legendaryCustomization: LegendaryCustomization? = null,
     val isLive: Boolean = false,
+    val nip05Status: Nip05VerificationStatus? = null,
```
AleksandarIlic: We already have `internetIdentifier` (which are nip05), and now we have `nip05Status`. Maybe we should call this `internetIdentifierStatus` or  `identifierStatus` and they should be next to each other. The reason why I did not wanted to name it `nip05` in the first place was because nip05 is just an implementation detail imo.

This comment applies to every ui component which has `nip05Status` along with `internetIdentifier`. 

I'm not sure what is the best way forward here, because this would apply also to the domain class name `Nip05VerificationStatus` as well.

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/core/compose/NostrUserText.kt
markocic: Yes. Verification service will exit early if the check was redundant. We need to re-verify after some time, or re-check failed verification etc.

data/caching/repository/src/commonMain/kotlin/net/primal/data/repository/factory/CommonRepositoryFactory.kt
markocic: This is according to nip. They explicitly state that we shouldn't follow redirects.

AleksandarIlic (approved):

PR #997, opened 2026-03-27
Add audio player for audio links in feed

mehmedalijaK:


app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteAudioPlayerPreview.kt
```diff
@@ -0,0 +1,309 @@
...
+private const val SECONDS_PER_HOUR = 3600
+
+@Composable
+fun NoteAudioPlayerPreview(
+    modifier: Modifier = Modifier,
+    eventId: String,
+    title: String?,
+    url: String,
+) {
+    val audioState = rememberAudioPlayerState(mediaId = eventId)
```
markocic: Having multiple media items in a single post breaks this logic. They seem to share same play/pause state.

app/src/main/kotlin/net/primal/android/core/video/MediaControllerExt.kt
```diff
@@ -76,6 +81,96 @@ fun rememberManagedMediaController(
     return controller
 }
 
+private const val PROGRESS_POLL_INTERVAL_MS = 200L
+
+@Composable
+fun rememberAudioPlayerState(mediaId: String): AudioPlayerState {
```
markocic: This player should stop once we open a video and continue playing if we exit. We have solved that in `StreamState` with a locking mechanism. Same pattern should be applied here. I think it would be best to make `AudioPlayerState` a stable class and all logic lives there. This `rememberAudioPlayerState` would just be a composable wrapper. Investigate how we have done that with StreamState.

app/src/main/kotlin/net/primal/android/core/video/MediaControllerExt.kt
```diff
@@ -76,6 +81,96 @@ fun rememberManagedMediaController(
...
+    var isPlaying by remember(mediaId) { mutableStateOf(false) }
+    var playWhenReady by remember(mediaId) { mutableStateOf(false) }
+    var isBuffering by remember(mediaId) { mutableStateOf(false) }
+    var currentPositionMs by remember(mediaId) { mutableLongStateOf(0L) }
+    var durationMs by remember(mediaId) { mutableLongStateOf(0L) }
+
+    DisposableEffect(mediaId) {
+        val token = SessionToken(
+            appContext,
+            ComponentName(appContext, PrimalMediaSessionService::class.java),
```
markocic: This will start a new Media Session for each audio. Streams use shared mediaController across multiple streams.

Also, are we fine with using `PrimalMediaSessionService`? I see some stream specific stuff there. When we click notification we should open the note that is playing this audio. Right now we are trying to launch a stream with this id which ultimately fails.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/AudioPlayerState.kt
```diff
@@ -0,0 +1,17 @@
...
+    val playWhenReady: Boolean = false,
+    val isBuffering: Boolean = false,
+    val progress: Float = 0f,
+    val currentPositionMs: Long = 0L,
+    val durationMs: Long = 0L,
+    val isActiveForMediaId: Boolean = false,
+    val play: () -> Unit = {},
+    val pause: () -> Unit = {},
+    val seekTo: (Long) -> Unit = {},
+    val playMediaItem: (MediaItem) -> Unit = {},
```
markocic: Not sure if we should have this state component coupled with `MediaItem`. This is an external dependency that can be changed. We should save url or something like that. Building `MediaItem` off of it.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteAudioPlayerPreview.kt
```diff
@@ -0,0 +1,309 @@
...
+            CircularProgressIndicator(
+                modifier = Modifier.size(20.dp),
+                strokeWidth = 2.dp,
+                color = AppTheme.colorScheme.onSecondary,
+            )
+        } else {
+            val isShowingPause = audioState.playWhenReady
+            Icon(
+                modifier = Modifier.size(if (isShowingPause) 32.dp else 20.dp),
+                imageVector = if (isShowingPause) PrimalIcons.VideoPauseMini else PrimalIcons.Play,
```
markocic: Let's use `VideoPause`. This is a mini player version.

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteAudioPlayerPreview.kt
```diff
@@ -0,0 +1,309 @@
...
+    durationMs: Long,
+) {
+    Row(
+        modifier = Modifier.fillMaxWidth(),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        val displayTitle = title ?: url.substringAfterLast("/").substringBefore("?")
+        Text(
+            modifier = Modifier.weight(1f),
+            text = displayTitle,
```
markocic, line 199: Not sure if we should have this. Mostly it's just blossom hashes. Do we have some alternatives?

app/src/main/kotlin/net/primal/android/notes/feed/note/ui/NoteAudioPlayerPreview.kt
```diff
@@ -0,0 +1,309 @@
...
+                strokeWidth = 2.dp,
+                color = AppTheme.colorScheme.onSecondary,
+            )
+        } else {
+            val isShowingPause = audioState.playWhenReady
+            Icon(
+                modifier = Modifier.size(if (isShowingPause) 32.dp else 20.dp),
+                imageVector = if (isShowingPause) PrimalIcons.VideoPauseMini else PrimalIcons.Play,
+                contentDescription = null,
+                tint = AppTheme.colorScheme.onSecondary,
```
markocic: We should use `Color.White`. Black color on light theme looks weird.

markocic (changes requested):

AleksandarIlic (approved):

PR #1015, opened 2026-04-14
Implement feed editor with Advanced Search prepopulation

mehmedalijaK:


app/src/main/kotlin/net/primal/android/explore/asearch/AdvancedSearchScreen.kt
```diff
@@ -118,6 +121,17 @@ private fun AdvancedSearchScreen(
...
+        if (state.loading) {
+            Box(
+                modifier = Modifier
+                    .fillMaxSize()
+                    .padding(paddingValues),
+                contentAlignment = Alignment.Center,
+            ) {
+                CircularProgressIndicator()
+            }
+            return@PrimalScaffold
```
AleksandarIlic: We never use this in codebase before. Please fix this!

app/src/main/kotlin/net/primal/android/explore/asearch/AdvancedSearchViewModel.kt
```diff
@@ -107,6 +118,58 @@ class AdvancedSearchViewModel @Inject constructor(
...
+                .onSuccess { parsed -> applyParsedQuery(parsed) }
+                .onFailure { Napier.w(throwable = it) { "Failed to parse advanced search query." } }
+            setState { copy(loading = false) }
+        }
+
+    private suspend fun applyParsedQuery(parsed: AdvancedSearchParsedQuery) {
+        val includedWords = buildString {
+            if (parsed.includes.isNotEmpty()) append(parsed.includes)
+            if (parsed.hashtags.isNotEmpty()) {
+                val tags = parsed.hashtags.split(" ").joinToString(" ") { "#$it" }
```
AleksandarIlic: Are you sure this is not going to produce empty `#` values like this? Maybe we should use something like:
`split("\\s+".toRegex()).filter { it.isNotBlank() }`

app/src/main/kotlin/net/primal/android/explore/feed/ExploreFeedViewModel.kt
```diff
@@ -46,6 +47,8 @@ class ExploreFeedViewModel @Inject constructor(
         ?: savedStateHandle.advancedSearchFeedSpec?.buildAdvancedSearchFeedSpec()
         ?: error("no feed spec provided.")
 
+    private val editingFeedSpec = savedStateHandle.editingFeedSpec
+
```
AleksandarIlic, line 53: Why do we have two feed specs here? Maybe we can use the original `feedSpec` and pass some other flag to indicate this is existing feed spec? WDYT?

app/src/main/kotlin/net/primal/android/explore/asearch/AdvancedSearchViewModel.kt
```diff
@@ -107,6 +118,58 @@ class AdvancedSearchViewModel @Inject constructor(
             }
         }
 
+    private fun parseEditingFeedSpec() =
+        viewModelScope.launch {
+            val query = editingFeedSpec?.extractAdvancedSearchQuery() ?: return@launch
+            setState { copy(loading = true) }
+            runCatching { feedsRepository.getAdvancedSearchQuery(query = query) }
+                .onSuccess { parsed -> applyParsedQuery(parsed) }
+                .onFailure { Napier.w(throwable = it) { "Failed to parse advanced search query." } }
```
AleksandarIlic: I'm not sure if this is safe to ignore? We were unable to parse feed spec, user could make a change on the default empty spec and that would override the existing feed spec being lost forever.

app/src/main/kotlin/net/primal/android/explore/asearch/AdvancedSearchViewModel.kt
```diff
@@ -107,6 +118,58 @@ class AdvancedSearchViewModel @Inject constructor(
             }
         }
 
+    private fun parseEditingFeedSpec() =
+        viewModelScope.launch {
+            val query = editingFeedSpec?.extractAdvancedSearchQuery() ?: return@launch
+            setState { copy(loading = true) }
+            runCatching { feedsRepository.getAdvancedSearchQuery(query = query) }
+                .onSuccess { parsed -> applyParsedQuery(parsed) }
```
AleksandarIlic: We should put this in mapCatching or something like that. If `applyParsedQuery` throws exception we are not catching that.

app/src/main/kotlin/net/primal/android/explore/feed/ExploreFeedViewModel.kt
```diff
@@ -96,7 +113,10 @@ class ExploreFeedViewModel @Inject constructor(
         viewModelScope.launch {
             events.collect {
                 when (it) {
-                    is UiEvent.AddToUserFeeds -> addToMyFeeds(it)
+                    is UiEvent.AddToUserFeeds -> {
+                        setState { copy(feedTitle = it.title, feedDescription = it.description) }
```
AleksandarIlic: This optimistic change needs to be reverted if `addToMyFeeds(it)` fails. We should probably handle all of this inside of  `addToMyFeeds`.

app/src/main/kotlin/net/primal/android/explore/asearch/AdvancedSearchScreen.kt
AleksandarIlic: Also, Search in Bottom bar is also visible while loading is happening. On slow connections, users will see Search button and be able to tap on it.

AleksandarIlic (changes requested):

AleksandarIlic (dismissed):

AleksandarIlic (approved):

PR #1023, opened 2026-04-21
Optimize account creation in onboarding flow

markocic:


app/src/main/kotlin/net/primal/android/auth/onboarding/account/OnboardingContract.kt
```diff
@@ -19,6 +19,7 @@ interface OnboardingContract {
         val followPacks: List<OnboardingFollowPack> = emptyList(),
         val expandedPackNames: Set<String> = emptySet(),
         val followedUserIds: Set<String> = emptySet(),
+        val defaultRelays: List<String>? = null,
```
AleksandarIlic: Why do we need this in state?

app/src/main/kotlin/net/primal/android/auth/onboarding/account/OnboardingViewModel.kt
```diff
@@ -83,6 +84,17 @@ class OnboardingViewModel @Inject constructor(
             setState { copy(working = false) }
         }
 
+    private fun fetchDefaultRelays() =
+        viewModelScope.launch {
+            runCatching { onboardingRepository.fetchDefaultRelays() }
+                .onSuccess { relays ->
+                    setState { copy(defaultRelays = relays) }
```
AleksandarIlic: I think this should be stored as private val in this viewModel and not in the state.

app/src/main/kotlin/net/primal/android/user/repository/RelayRepository.kt
```diff
@@ -34,13 +34,17 @@ class RelayRepository @Inject constructor(
     fun findRelays(userId: String, kind: RelayKind) = usersDatabase.relays().findRelays(userId, kind)
 
     @Throws(NostrPublishException::class, SignatureException::class)
-    suspend fun bootstrapUserRelays(userId: String) =
+    suspend fun bootstrapUserRelays(userId: String, preFetchedRelayUrls: List<String>? = null) =
```
AleksandarIlic: This function has two responsibilities now. Can we split this into two public function with clear responsibilities?

app/src/main/kotlin/net/primal/android/auth/repository/CreateAccountHandler.kt
```diff
@@ -38,20 +46,46 @@ class CreateAccountHandler @Inject constructor(
...
+            coroutineScope {
+                awaitAll(
+                    async { blossomRepository.ensureBlossomServerList(userId) },
+                    async { userRepository.setProfileMetadata(userId = userId, profileMetadata = profileMetadata) },
+                    async { userRepository.setFollowList(userId = userId, contacts = setOf(userId) + followedUserIds) },
+                )
+            }
+
+            scope.launch { initWallet(userId) }
+            scope.launch { fetchSettings(userId) }
```
AleksandarIlic: We should put withTimeout to both of the coroutines to make sure this is not taking too much time since scope is never canceled. Let's put 30.seconds.

AleksandarIlic (changes requested):

app/src/main/kotlin/net/primal/android/auth/repository/CreateAccountHandler.kt
```diff
@@ -38,20 +46,46 @@ class CreateAccountHandler @Inject constructor(
...
+
+            coroutineScope {
+                awaitAll(
+                    async { blossomRepository.ensureBlossomServerList(userId) },
+                    async { userRepository.setProfileMetadata(userId = userId, profileMetadata = profileMetadata) },
+                    async { userRepository.setFollowList(userId = userId, contacts = setOf(userId) + followedUserIds) },
+                )
+            }
+
+            scope.launch { initWallet(userId) }
```
AleksandarIlic: Can we run these two at beginning so this kicks in witout waiting for scope above to finsih?

AleksandarIlic (changes requested):

AleksandarIlic (approved):

PR #1030, opened 2026-04-23
Fix DVM feed bottom sheet in new Explore

mehmedalijaK:


app/src/main/kotlin/net/primal/android/feeds/dvm/ui/DvmFeedDetailsBottomSheet.kt
```diff
@@ -0,0 +1,146 @@
...
+import net.primal.android.core.compose.PrimalScaffold
+import net.primal.android.core.compose.icons.PrimalIcons
+import net.primal.android.core.compose.icons.primaliconpack.ArrowBack
+import net.primal.android.core.errors.UiError
+import net.primal.android.notes.feed.note.ui.events.NoteCallbacks
+import net.primal.android.theme.AppTheme
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DvmFeedDetailsBottomSheet(
```
markocic, line 35: We should delete old private one. No need to have duplicate code.

markocic (changes requested):

markocic (approved):

PR #1035, opened 2026-04-24
Replace deprecated LocalClipboardManager with copyText and pasteText helpers

mehmedalijaK:


app/src/main/kotlin/net/primal/android/core/utils/ClipboardUtils.kt
```diff
@@ -22,6 +22,11 @@ fun copyText(
     clipboard.setPrimaryClip(clip)
 }
 
+fun pasteText(context: Context): String {
```
AleksandarIlic: Maybe `Context.pasteText()` would be better?

AleksandarIlic (approved):

PR #1036, opened 2026-04-24
Chunk IN-clause DAO queries to avoid SQLite variable-limit crash on older devices

markocic:


AleksandarIlic (approved):
Very nice solution! Great work!

PR #1037, opened 2026-04-27
Redesign Explore tab header with filter picker and search bar

mehmedalijaK:


app/src/main/kotlin/net/primal/android/main/explore/filter/ExploreFilterListOverlayContent.kt
```diff
@@ -0,0 +1,117 @@
...
+import androidx.compose.ui.unit.sp
+import net.primal.android.theme.AppTheme
+
+@Composable
+fun ExploreFilterListOverlayContent(
+    activeFilter: ExploreFilter,
+    onFilterClick: (ExploreFilter) -> Unit,
+    onDismiss: () -> Unit,
+) {
+    BackHandler { onDismiss() }
```
markocic: We have this in PrimalOverlay already. No need for it.

app/src/main/kotlin/net/primal/android/main/explore/filter/ExploreFilterListOverlayContent.kt
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+}
+
+@Composable
+private fun ExploreFilterListItem(
+    modifier: Modifier,
+    filter: ExploreFilter,
+    selected: Boolean,
+) {
```
markocic: Don't we have component for something like this already for feed picker? Could we make that one more general and reuse it here as well? They will likely follow similar design in the future.

app/src/main/kotlin/net/primal/android/main/explore/filter/ExploreFilter.kt
```diff
@@ -0,0 +1,30 @@
+package net.primal.android.main.explore.filter
+
+import androidx.annotation.StringRes
+import net.primal.android.R
+
+enum class ExploreFilter(
+    @StringRes val titleRes: Int,
+    @StringRes val subtitleRes: Int,
+) {
```
markocic: Let's go with enum and two extension functions. `ExploreFilter.toTitle()` and `ExploreFilter.toSubtitle()`. I think these are established patterns for cases like this.

app/src/main/kotlin/net/primal/android/main/explore/filter/ExploreFilter.kt
```diff
@@ -0,0 +1,30 @@
+package net.primal.android.main.explore.filter
+
+import androidx.annotation.StringRes
+import net.primal.android.R
+
+enum class ExploreFilter(
```
markocic: Not sure about word Filter. `ExploreTab` could be better, open to suggestions. Make sure to change everywhere.

app/src/main/kotlin/net/primal/android/main/MainScreen.kt
```diff
@@ -389,15 +406,15 @@ private fun MainScreenContent(
                         navController = navController,
                     )
 
-                    PrimalTopLevelDestination.Explore -> NewExploreTabContent(
+                    PrimalTopLevelDestination.Explore -> ExploreHomeContent(
```
markocic, line 416: When scrolling down we shouldn't hide top and bottom app bar.

app/src/main/kotlin/net/primal/android/main/explore/ExploreHomeContent.kt
```diff
@@ -56,89 +64,86 @@ internal fun ExploreHomeContent(
...
-        when (pageIndex) {
-            PEOPLE_INDEX -> {
-                ExplorePeople(
-                    modifier = Modifier.background(color = AppTheme.colorScheme.surfaceVariant),
-                    paddingValues = paddingValues,
-                    onProfileClick = { noteCallbacks.onProfileClick?.invoke(it) },
-                    onFollowPackClick = onFollowPackClick,
-                )
-            }
+    when (activeFilter) {
```
markocic: This should be horizontally scrollable, just like feeds are. Also make sure that we keep scroll animation, same as feeds.

markocic (changes requested):

app/src/main/kotlin/net/primal/android/main/MainScreen.kt
```diff
@@ -452,6 +470,7 @@ private fun rememberMainScreenSharedState(
         snackbarHostState = remember { SnackbarHostState() },
         homeActiveFeed = remember { mutableStateOf(null) },
         readsActiveFeed = remember { mutableStateOf(null) },
+        exploreActiveFilter = remember { mutableStateOf(ExploreFilter.Explore) },
```
markocic: Let's go with `rememberSaveable`.

app/src/main/res/values/strings.xml
```diff
@@ -720,15 +720,23 @@
     <string name="new_explore_user_search_placeholder">Search users</string>
     <string name="new_explore_fab_content_description">Open new Explore</string>
 
+    <!-- Explore tab header picker -->
+    <string name="explore_filter_explore_title">Explore</string>
+    <string name="explore_filter_explore_subtitle">Find people, feeds, media</string>
+    <string name="explore_filter_feed_gallery_title">Feed Gallery</string>
+    <string name="explore_filter_feed_gallery_subtitle">Browse feed gallery</string>
+    <string name="explore_filter_follow_packs_title">Follow Packs</string>
+    <string name="explore_filter_follow_packs_subtitle">Created lists of users to follow</string>
```
markocic: It should say Curated. Check the rest to make it's the same.

markocic (changes requested):

markocic (approved):

PR #1058, opened 2026-06-24
Fix parseFollowings to validate public keys from p tags

akurcubic:


app/src/main/kotlin/net/primal/android/user/accounts/Parsers.kt
```diff
@@ -48,7 +51,9 @@ fun List<JsonArray>.parseFollowings(): Set<String> {
     this.forEach {
         if (it.getOrNull(0)?.jsonPrimitive?.content == "p") {
             val pubkey = it.getOrNull(1)?.jsonPrimitive?.content
-            if (pubkey != null) followings.add(pubkey)
+            if (pubkey != null && pubkey.length == HEX_PUBKEY_LENGTH && pubkey.isValidHex()) {
```
AleksandarIlic: `pubkey.length == HEX_PUBKEY_LENGTH` should be moved into`isValidHex()`

AleksandarIlic (changes requested):
Looks great! Please fix minor comment before we can merge this.

AleksandarIlic (approved):

PR #1064, opened 2026-06-29
Apply keepLoaded() to the Flow<PagingData> sources

akurcubic:


AleksandarIlic (approved):
With collectLatest we can get rid of Jobs.

PR #1069, opened 2026-07-01
Hoist zapping state into app scoped CompositionLocal

andpantic:


app/src/main/kotlin/net/primal/android/core/activity/PrimalActivity.kt
```diff
@@ -119,3 +129,5 @@ val LocalPrimalTheme = compositionLocalOf<PrimalTheme> { error("No PrimalTheme p
 val LocalContentDisplaySettings = compositionLocalOf<ContentDisplaySettings> {
     error("No ContentDisplay settings provided.")
 }
+
+val LocalZappingState = compositionLocalOf { ZappingState() }
```
markocic: nitpick: We could do `error` like we do with other providers above. Idea is for application to crash instead of providing false value on runtime due to programmer error in the future.

markocic (dismissed):

AleksandarIlic (approved):

PR #1078, opened 2026-07-03
NIP-22 article comments (kind 1111)

andpantic:


app/src/main/kotlin/net/primal/android/editor/NotePublishHandler.kt
```diff
@@ -265,6 +295,38 @@ class NotePublishHandler @Inject constructor(
...
+        articleEventId: String?,
+        replyTo: Nevent?,
+    ): List<JsonArray> =
+        buildList {
+            add(article.asRootReplaceableEventTag())
+            add(NostrEventKind.LongFormContent.asRootKindTag())
+            add(article.asRootPubkeyTag())
+            if (replyTo != null) {
+                add(replyTo.asCommentEventTag())
+                add(NostrEventKind.Comment.asKindTag())
```
markocic: We should go with `replyTo.kind` here as this could be a reply to `kind: 1` comment that was left previously and we would be referencing event that doesn't exist.

markocic (changes requested):

PR #1079, opened 2026-07-03
Feature/room db space cleanup

markocic:


AleksandarIlic:
Let's better not introduce risks with WAL shrinking since it's time-consuming. I have applied `858cabd14` config options for fine-tuning per database and additionally I'll do database rename for releasing old storage. Wallet/Account db fixes should be done in separate PR.

PR #1081, opened 2026-07-06
Reserve space for Followed by section on profile screen

akurcubic:


app/src/main/kotlin/net/primal/android/profile/details/ProfileDetailsViewModel.kt
```diff
@@ -241,6 +241,8 @@ class ProfileDetailsViewModel @Inject constructor(
                 }
             } catch (error: NetworkException) {
                 Napier.e(throwable = error) { "Failed to fetch profiles followed by" }
+            } finally {
+                setState { copy(isResolvingFollowedBy = false) }
```
AleksandarIlic, line 246: Set `isResolvingFollowedBy = true` at the start so it's safe to call this function any time.

AleksandarIlic (changes requested):

AleksandarIlic (approved):

